PR #3, opened 2020-02-28
Fix references

habbes:
- Add missing dependencies for OData v3.
- Update root dir in the build pipeline

I added the binaries from the original connected service repo. I tried to add the dependencies as NuGet package but the build failed.


habbes:
@paulodero 

> Do we need to maintain these libraries in the repo? Or these could be restored in the pipeline before the build?

As mention in the description, I tried to get them from NuGet so that they can be restored, but that still failed to build. Maybe it's because the versions were old, I'm not sure. Still got unresolved reference errors. I assume that's why the previous repo stored them as binaries. The Unchase version also does this.

paulodero:
> @paulodero
> 
> Do we need to maintain these libraries in the repo? Or these could be restored in the pipeline before the build?
> 
> As mention in the description, I tried to get them from NuGet so that they can be restored, but that still failed to build. Maybe it's because the versions were old, I'm not sure. Still got unresolved reference errors. I assume that's why the previous repo stored them as binaries. The Unchase version also does this.

Have you tried doing the same in VS to see where the challenge is? It would be great to have these restored in the build pipeline.

habbes:
@paulodero 
> Have you tried doing the same in VS to see where the challenge is? It would be great to have these restored in the build pipeline.

Yes, I have tried doing the same in VS. I have tried adding references, tried adding Nuget packages, etc. They did not work. Adding global references depends on the packages being already installed in your system (GAC), which is not something you can rely on. I know it would be preferrable not to include the binaries, but in the absence of viable alternatives, I don't think it's a big problem, especially given that they are only there for backwards compatibility (Odata v3 support) and we won't need to update them.

PR #5, opened 2020-03-02
Fix outdated template file to match code generator

habbes:
The template file in the codebase (`ODataT4CodeGenerator.ttinclude`) was incorrect, it's an outdated version from an old commit that added the `internal` modifier to both method and classes when `MakeTypesInternal` is set to true. However, this generated code that had build errors because internal methods can't implement interfaces (e.g. `INotifyChanged`), this was fixed by a later commit. But it appears that during a rebase, the outdated template fail was retained instead of the corrected one. This PR fixes that issue, removing the internal class modifier from methods that should be public.

Note the `ODataT4CodeGenerator.cs` file was still correct, and that's why tests were still passing.

### DIFF
--- a/src/Templates/ODataT4CodeGenerator.cs
+++ b/src/Templates/ODataT4CodeGenerator.cs
@@ -333,8 +333,8 @@ public bool IgnoreUnexpectedElementsAndAttributes
 /// </summary>
 public bool MakeTypesInternal
 {
-	get;
-	set;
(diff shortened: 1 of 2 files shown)


habbes:
So just to be clear, the `MakeTypesInternal` option should only affect types, i.e. classes, enums and the static class that would have otherwise been made public.

PR #6, opened 2020-03-02
Convert type definitions to the underlying type

habbes:
Fixes issue https://github.com/OData/lab/issues/81

The connected service could not generate code for the Outlook Beta API mentioned in the related issue due to 2 main issues:

### Issue 1
The outlook beta API mentioned in the related issue contained the following type definition:
```xml
<TypeDefinition Name="DateTime" UnderlyingType="Edm.String" />
```
Apparently the code generator did not know how to handle such cases. It tried to generate its code as an `IEdmEntityType`, and that lead to null-reference exceptions. Similar issues occurred when translating function parameters whose types are based on type definitions.

This PR solves this issue in two steps:
1. Ignore type definitions when writing classes/enums for entities, complex types and enums
2. When writing function parameters, use the underlying primitive type of the type definition as the parameter type.
E.g: a function defined in the schema as:
```xml
<Function Name="MyFunc" IsBound="true">
  <Parameter Name="StartDate" Type="MyNamespace.DateTime" />
  <ReturnType Type="Edm.Int32" Nullable="false" />
</Function>
```
will be defined in the generated C# code as:
```c#
public int MyFunc(string StartDate)
{
//...
}
```
Note that using `MyNamespace.DateTime` as a type in the C# code would cause compile-time errors because there's no equivalent C# type generated for the type definition by the code generator.

### Issue 2
For edm types like `TimeOfDay` which are defined in the OData Edm library, the connected service was generating code in the wrong namespace, i.e. generating `Microsoft.OData.Edm.Library.TimeOfDay` instead of `Microsoft.OData.Edm.TimeOfDay`. It was probably generating code for an older version of the lib. The code generator itself uses an older version of edm lib (6.x). We should update it to 7.x soon.


habbes:
Due to the changes brought to the repo after the migration effort from code gen, it became more manageable to re-create this fix on a new branch rather than update this PR. The new PR is at: https://github.com/OData/ODataConnectedService/pull/25

Issue #11, opened 2020-03-04
Rename EdmxReaderSettings to CsdlReaderSettings


KenitoInc:
In OData v4, we use `CsdlReaderSettings` instead of `EdmxReaderSettings`

PR #13, opened 2020-03-04
Rename EdmxReaderSettings to CsdlReaderSettings

KenitoInc:
This fixes Issue #11


src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// ------------------------------------------------------------------------------
 // <auto-generated>
 //     This code was generated by a tool.
-//     Runtime Version: 16.0.0.0
+//     Runtime Version: 15.0.0.0
```
paulodero: We had decided to leave this at version 16. You could use VS 2019 to ensure this version is maintained at 16...

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -31,7 +31,7 @@ namespace Microsoft.OData.ConnectedService.Templates
     /// <summary>
     /// Class to produce the template output
     /// </summary>
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.TextTemplating", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.TextTemplating", "15.0.0.0")]
```
paulodero: VS version

PR #19, opened 2020-03-06
Upgrade Microsoft.OData.Client lib to version 7.4.0

paulodero:
Upgrading `Microsoft.OData.Client` lib to version 7.4.0 to facilitate migration of code gen. The intention is to finally have the latest version 7.6.3 but to make migration easier, upgrade is being done in stages.

### DIFF
--- a/src/Templates/ODataT4CodeGenerator.cs
+++ b/src/Templates/ODataT4CodeGenerator.cs
@@ -19,10 +19,7 @@ namespace Microsoft.OData.ConnectedService.Templates
     using System.Collections.Generic;
     using Microsoft.OData.Edm.Csdl;
     using Microsoft.OData.Edm;
-    using Microsoft.OData.Edm.Annotations;
-    using Microsoft.OData.Edm.Expressions;
-    using Microsoft.OData.Edm.Library;
-    using Microsoft.OData.Edm.Values;
+    using Microsoft.OData.Edm.Vocabularies;
     using Microsoft.OData.Edm.Vocabularies.V1;
     using Microsoft.OData.Edm.Vocabularies.Community.V1;
     using System.Text;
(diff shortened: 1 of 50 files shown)


src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -626,12 +623,12 @@ public IEdmModel EdmModel
                 Debug.Assert(this.edmx != null, "this.edmx != null");
 
                 IEnumerable<Microsoft.OData.Edm.Validation.EdmError> errors;
-                EdmxReaderSettings edmxReaderSettings = new EdmxReaderSettings()
+                CsdlReaderSettings edmxReaderSettings = new CsdlReaderSettings()
```
KenitoInc: Change `edmxReaderSettings` to `csdlReaderSettings`

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -514,12 +511,12 @@ public class CodeGenerationContext
                 Debug.Assert(this.edmx != null, "this.edmx != null");
 
                 IEnumerable<Microsoft.OData.Edm.Validation.EdmError> errors;
-                EdmxReaderSettings edmxReaderSettings = new EdmxReaderSettings()
+                CsdlReaderSettings edmxReaderSettings = new CsdlReaderSettings()
```
KenitoInc: Change edmxReaderSettings to csdlReaderSettings

src/Templates/ODataT4CodeGenerator.cs
paulodero: Fixed

src/Templates/ODataT4CodeGenerator.ttinclude
paulodero: Fixed

PR #21, opened 2020-03-06
Add semmle task to pipeline

habbes:
**NOTE: Not ready to merge**

- Adds the [Semmle](https://semmle.com/) static analyzer to the pipeline to detect security vulnerabilities
- Since Semmle run is time consuming (takes about 16min alone, causes entire build to run for 20+min), it's configured to only run when on `master` or `release` branches, so it will be skipped for commits made against a PR branch
- Semmle is also run when the build is triggered by a schedule (e.g. nightly build)
- Configured a nightly build to run every day at 3AM UTC (6AM EAT)

I've temporarily added this branch (`integrate-semmle`) to the nightly builds so that we see whether the builds and semmle run as expected before we decide to merge.


odero:
@habbes this should either be merged or closed. Rebase to master if it's to be merged

habbes:
The Semmle task doesn't seem to be available in the new pipeline. I'm not sure if it was discontinued or whether we need to install it. But I have not found it in the VS Marketplace.

PR #22, opened 2020-03-09
Use CsdlReader.TryParse(..) instead of CsdlReader.Parse(..)

paulodero:
This pull request is part of codegen/connected service migration. makes use of `CsdlReader.TryParse(..)` instead of `CsdlReader.Parse(...)`.

### DIFF
--- a/src/Templates/ODataT4CodeGenerator.cs
+++ b/src/Templates/ODataT4CodeGenerator.cs
@@ -4007,7 +4007,25 @@ this.Write("                    global::System.Xml.XmlReader reader = CreateXmlR
 
 this.Write(@"                try
                 {
-                    return global::Microsoft.OData.Edm.Csdl.CsdlReader.Parse(reader);
+                    global::System.Collections.Generic.IEnumerable<global::Microsoft.OData.Edm.Validation.EdmError> errors;
+                    global::Microsoft.OData.Edm.IEdmModel edmModel;
+
+                    if (!global::Microsoft.OData.Edm.Csdl.CsdlReader.TryParse(reader, ");
+
+this.Write(this.ToStringHelper.ToStringWithCulture(this.context.IgnoreUnexpectedElementsAndAttributes ? "true" : "false"));
+
+this.Write(@", out edmModel, out errors))
+                    {
+	                    global::System.Text.StringBuilder errorMessages = new System.Text.StringBuilder();
+	                    foreach (var error in errors)
+	                    {
+		                    errorMessages.Append(error.ErrorMessage);
+		                    errorMessages.Append(""; "");
+	                    }
+	                    throw new global::System.InvalidOperationException(errorMessages.ToString());
+                    }
+
+        return edmModel;
                 }
                 finally
                 {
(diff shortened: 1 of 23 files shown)


paulodero:
> Parse()



> @paulodero  Out of curiosity, what is the advantage of using TryParse() instead of Parse() if we are still just throwing an exception. I assume Parse() already throws an exception if the parse fails.

@habbes with `TryParse()` all error messages are concatenated into one message and this forms the error message for `InvalidOperationException()` being thrown. I believe this was the intention for this change in Code Gen.

habbes:
@paulodero 
> @habbes with `TryParse()` all error messages are concatenated into one message and this forms the error message for `InvalidOperationException()` being thrown. I believe this was the intention for this change in Code Gen.

I see

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -4007,7 +4007,25 @@ internal override void WriteGeneratedEdmModel(string escapedEdmxString)
 
 this.Write(@"                try
                 {
-                    return global::Microsoft.OData.Edm.Csdl.CsdlReader.Parse(reader);
+                    global::System.Collections.Generic.IEnumerable<global::Microsoft.OData.Edm.Validation.EdmError> errors;
+                    global::Microsoft.OData.Edm.IEdmModel edmModel;
+
+                    if (!global::Microsoft.OData.Edm.Csdl.CsdlReader.TryParse(reader, ");
+
+this.Write(this.ToStringHelper.ToStringWithCulture(this.context.IgnoreUnexpectedElementsAndAttributes ? "true" : "false"));
```
odero: Maybe use the syntax you have in the ttinclude - it's easier to read.

test/ODataConnectedService.Tests/CodeGenReferences/AbstractEntityTypeWithoutKeyDSC.cs
```diff
@@ -166,7 +166,21 @@ private abstract class GeneratedEdmModel
...
+                    if (!global::Microsoft.OData.Edm.Csdl.CsdlReader.TryParse(reader, false, out edmModel, out errors))
+                    {
+                        global::System.Text.StringBuilder errorMessages = new System.Text.StringBuilder();
+                        foreach (var error in errors)
+                        {
+                            errorMessages.Append(error.ErrorMessage);
+                            errorMessages.Append("; ");
+                        }
+                        throw new global::System.InvalidOperationException(errorMessages.ToString());
+                    }
```
odero: Keep things DRY: Seeing as this block of code is repeated in 20+ test files it would be worth considering extracting it to a separate location (e.g. a helper file) where it can be called by the 20+ files instead of duplicating it

src/Templates/ODataT4CodeGenerator.cs
paulodero: This is the `.cs `file generated by `.tt `used to generate text when using runtime t4 templates. The content can not be the same with what is in `.ttinclude`. Actually this file is autogenerated.

test/ODataConnectedService.Tests/CodeGenReferences/AbstractEntityTypeWithoutKeyDSC.cs
paulodero: These test files are resource files. They are not compiled or executed. Their content is only used to compare generated code. Not sure if is possible to write a helper file for an `embedded resource `in a project. Any idea @g2mula @gathogojr ?

src/Templates/ODataT4CodeGenerator.cs
odero: My bad. I'd thought it was the .tt file. Ignore

PR #23, opened 2020-03-09
Apply fix from CodeGen: customize built-in vocab models

habbes:
- Original issue: https://github.com/OData/odata.net/issues/1285
- Original commit: https://github.com/OData/odata.net/commit/8ee92c44
- Original PR: https://github.com/OData/odata.net/pull/1367

### DIFF
--- a/src/Templates/ODataT4CodeGenerator.cs
+++ b/src/Templates/ODataT4CodeGenerator.cs
@@ -1075,7 +1075,9 @@ public class CodeGenerationContext
                 tmp.FindDeclaredTerm(CoreVocabularyConstants.OptimisticConcurrency) != null ||
                 tmp.FindDeclaredTerm(CapabilitiesVocabularyConstants.ChangeTracking) != null ||
                 tmp.FindDeclaredTerm(AlternateKeysVocabularyConstants.AlternateKeys) != null ||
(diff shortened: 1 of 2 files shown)


src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1075,7 +1075,9 @@ private static IEnumerable<T> GetElementsFromModelTree<T>(IEdmModel mainModel, F
                 tmp.FindDeclaredTerm(CoreVocabularyConstants.OptimisticConcurrency) != null ||
                 tmp.FindDeclaredTerm(CapabilitiesVocabularyConstants.ChangeTracking) != null ||
                 tmp.FindDeclaredTerm(AlternateKeysVocabularyConstants.AlternateKeys) != null ||
-                tmp.FindDeclaredTerm("Org.OData.Authorization.V1.Authorizations") != null)
+                tmp.FindDeclaredTerm("Org.OData.Authorization.V1.Authorizations") != null ||
```
paulodero: Was this change necessary?

habbes: Which change? The additional reference models?

paulodero: I get it.

PR #26, opened 2020-03-10
Auto open generated files in the IDE when generation completes support

ElizabethOkerio:
This PR adds support for automatically opening the generated files in the IDE when generation of the files completes.

### DIFF
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -20,7 +20,7 @@ namespace Microsoft.OData.ConnectedService.Models
         [DataMember]
         public ObservableCollection<string> MruEndpoints { get; private set; }
         [DataMember]
-        public bool OpenGeneratedFilesOnIDE { get; set; } = false;
+        public bool OpenGeneratedFilesInIDE { get; set; } = false;
 
         private UserSettings()
         {
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -19,8 +19,6 @@ namespace Microsoft.OData.ConnectedService.Models
 
         [DataMember]
         public ObservableCollection<string> MruEndpoints { get; private set; }
-        [DataMember]
-        public bool OpenGeneratedFilesInIDE { get; set; } = false;
 
         private UserSettings()
         {
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -19,6 +19,8 @@ namespace Microsoft.OData.ConnectedService.Models
 
         [DataMember]
         public ObservableCollection<string> MruEndpoints { get; private set; }
+        [DataMember]
+        public bool OpenGeneratedFilesOnIDE { get; set; } = false;
 
         private UserSettings()
         {
--- a/src/Views/ConfigODataEndpoint.xaml
+++ b/src/Views/ConfigODataEndpoint.xaml
@@ -24,6 +24,5 @@
                   Margin="20, 5, 5, 0" IsEditable="True"
                   ItemsSource="{Binding Path=UserSettings.MruEndpoints}"
                   Text="{Binding Path=Endpoint, TargetNullValue='Enter your endpoint...' }"/>
-        
     </StackPanel>
 </UserControl>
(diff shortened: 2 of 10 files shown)


src/Views/ConfigODataEndpoint.xaml
```diff
@@ -24,5 +24,12 @@
                   Margin="20, 5, 5, 0" IsEditable="True"
                   ItemsSource="{Binding Path=UserSettings.MruEndpoints}"
                   Text="{Binding Path=Endpoint, TargetNullValue='Enter your endpoint...' }"/>
+        <CheckBox
+            x:Name="OpenGeneratedFilesOnIDE"
+            Margin="20,15,10,5"
+            HorizontalAlignment="Stretch"
+            Content="Open Generated Files In IDE When Generation Completes."
```
odero:
```suggestion
            Content="Open generated files in the IDE when generation completes."
```

src/Models/UserSettings.cs
```diff
@@ -19,6 +19,8 @@ internal class UserSettings
 
         [DataMember]
         public ObservableCollection<string> MruEndpoints { get; private set; }
+        [DataMember]
+        public bool OpenGeneratedFilesOnIDE { get; set; } = false;
```
odero:
```suggestion
        public bool OpenGeneratedFilesInIDE { get; set; } = false;
```

src/Views/ConfigODataEndpoint.xaml
```diff
@@ -24,5 +24,12 @@
                   Margin="20, 5, 5, 0" IsEditable="True"
                   ItemsSource="{Binding Path=UserSettings.MruEndpoints}"
                   Text="{Binding Path=Endpoint, TargetNullValue='Enter your endpoint...' }"/>
+        <CheckBox
+            x:Name="OpenGeneratedFilesInIDE"
+            Margin="20,15,10,5"
+            HorizontalAlignment="Stretch"
+            Content="Open generated files in the IDE when generation completes."
+            FontWeight="Bold"
+            IsChecked="{Binding Path=UserSettings.OpenGeneratedFilesInIDE, Mode=TwoWay}" />
```
paulodero: I feel this option should not be made available in the `Endpoint `tab but rather as an advanced settings

src/Views/ConfigODataEndpoint.xaml
```diff
@@ -24,5 +24,12 @@
                   Margin="20, 5, 5, 0" IsEditable="True"
                   ItemsSource="{Binding Path=UserSettings.MruEndpoints}"
                   Text="{Binding Path=Endpoint, TargetNullValue='Enter your endpoint...' }"/>
+        <CheckBox
+            x:Name="OpenGeneratedFilesInIDE"
+            Margin="20,15,10,5"
+            HorizontalAlignment="Stretch"
+            Content="Open generated files in the IDE when generation completes."
+            FontWeight="Bold"
```
paulodero: Any reason making this option bold (makes it look more important than other options which I do not think is the case)?

src/Views/ConfigODataEndpoint.xaml.cs
```diff
@@ -6,14 +6,19 @@
 
 namespace Microsoft.OData.ConnectedService.Views
 {
+
+    public delegate void OpenGeneratedFilesInIDE(object sender, RoutedEventArgs e);
```
odero:
```suggestion
    public delegate void OpenGeneratedFilesInIDEHandler(object sender, RoutedEventArgs e);
```
c# naming conventions

unchase:
Why this option allows only for OData v4 in `AdvancedSettings.xaml` (not for both v3 and v4 in `ConfigODataEndpoint.xaml`)?

ElizabethOkerio:
@unchase we thought it wise to add this option as a setting in <code> AdvancedSettings.xaml </code> and not part of the endpoint. The <code>
ConfigODataEndPoint.xaml> will only have settings that configure the endpoint.

unchase:
@ElizabethOkerio, in this case, you must enable the option for OData v 3 as well.

ElizabethOkerio:
@unchase. yes thanks will do that.

PR #27, opened 2020-03-10
Add support for Custom Http Headers

KenitoInc:
Part of codegen<>connected service parity
Original Issue: https://github.com/OData/odata.net/issues/1315
Original PR: https://github.com/OData/odata.net/pull/1637

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
 // Licensed under the MIT License.  See License.txt in the project root for license information.
 
+using System;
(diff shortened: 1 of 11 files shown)


KenitoInc:
@paulodero 
1. For simplicity, we have a multiline textbox where a user inputs http headers as strings e.g
```
headerKey:headerValue
headerKey:headerValue
headerKey:headerValue
````
We are trying to reduce complexicity in the ODataConnectedServiceWizard

2. This makes sense. We need to filter out http headers in the generated json file. Maybe for a start we can filter out everything (since it's not needed anyway)

KenitoInc:
> Is it possible to add a test that shows the headers from the user being used in the request?

Yes. On it

unchase:
Perhaps a description should be added to this option. It is completely incomprehensible to the user: what is it and in what context is it use?

PR #30, opened 2020-03-11
Generate multiple files

paulodero:
This is pull request allows connected service to generate multiple files.


src/Properties/AssemblyInfo.cs
```diff
@@ -39,4 +39,4 @@
 [assembly: AssemblyFileVersion("0.3.0.0")]
 [assembly: NeutralResourcesLanguageAttribute("en")]
 
-[assembly: InternalsVisibleTo("ODataConnectedService.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("ODataConnectedService.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
KenitoInc: Why are we updating the assembly info? This should never be updated since it's used by the build pipeline to run tests.
I think you deleted an empty line at the end of the file. Kindly fix this

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -24,6 +24,10 @@ namespace Microsoft.OData.ConnectedService.Templates
     using Microsoft.OData.Edm.Vocabularies.Community.V1;
     using System.Text;
     using System.Net;
+    using System.Security;
+    using Microsoft.VisualStudio.TextTemplating;
+    using Microsoft.OData.ConnectedService.CodeGeneration;
+    using Microsoft.VisualStudio.ConnectedServices;
```
marabooy: sort namespaces

src/Properties/AssemblyInfo.cs
paulodero: Do we really need the empty line?

src/Templates/ODataT4CodeGenerator.cs
paulodero: I though System directives should come first as above. What's your suggestion on this?

habbes:
@paulodero I get the following error when I try to generate client code for a service (TripPinRestierSevice) and check the "Generate multiple files" checkbox in the Advanced Settings. The error does not occur when I don't enable that option.

image

habbes:
@paulodero a follow-up to the error I reported. The error does not occur when you run Visual Studio as administrator. Is this expected behaviour? I would suggest looking into whether it's possible to add the feature without being an admin. And if there's not workaround, then adding a note next to the feature indicating that admin privileges are required.

paulodero:
> @paulodero a follow-up to the error I reported. The error does not occur when you run Visual Studio as administrator. Is this expected behaviour? I would suggest looking into whether it's possible to add the feature without being an admin. And if there's not workaround, then adding a note next to the feature indicating that admin privileges are required.

I am not running as admin yet I can generate code. Looks like you do not have rights in some of your directories. Let me look for work around for such cases.

src/Templates/ODataT4CodeGenerator.cs
marabooy: O>V for the microsoft ones

paulodero: Really?

habbes:
@paulodero when updating the connected service, the UI should be pre-populated with the settings that were saved in the json file. This is not the case for the "Generate multiple files" settings, so maybe you could add that too.

Actually, I think updating this setting could introduce some complexity. If the code was generated in multiple files, and then you update the connected service and disable the option to generate multiple files, then it will only generate a single file. This will lead to conflicts if you don't also delete all the extra files that were generated the first time. What I suggest is to simply disable this option once the code has been generated the first time for that service, similar to how the service endpoint and generated file name cannot be changed when you update the connected service.

I think there's also the case where the service itself might have removed or renamed an EntityType (for example), this would lead to the connected retaining an unused file. But I don't think that's a problem we should address in this PR, unless it ever occurs in the wild.

paulodero:
> I got an error when I tried to generate client code with the "Generate multiple files" setting enabled. The error does not occur when the setting is disabled.

@habbes Do you know why you are getting this error? I don't have it. Even if I run VS as a normal user and not the admin.

habbes:
> Do you know why you are getting this error? I don't have it. Even if I run VS as a normal user and not the admin.

@paulodero I don't fully understand the error other than it only occurred when I enabled multiple-file generation and disappeared when I used VS as an Admin. But looking at the error message, it seems that it's trying to access (not sure whether read or write) `Feature.cs` in the path where the IDE is located. `Feature.cs` is one of the generated files for the TripPinRestierService and it contains the definition of the `Feature` enum. So my guess is that it was trying to save the `Feature.cs` file at that location temporarily before moving it to its final destination. But if it was indeed just saving to a temp location, why would it pick the IDE path?

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -1167,9 +1215,12 @@ public abstract class ODataClientTemplate : TemplateBase
     /// <returns>The generated code for the OData client.</returns>
     public override string TransformText()
     {
+        context.MultipleFilesManager.StartHeader();
         this.WriteFileHeader();
+        context.MultipleFilesManager.EndBlock();
         this.WriteNamespaces();
-        return this.GenerationEnvironment.ToString();
+        context.MultipleFilesManager.GenerateFiles(context.GenerateMultipleFiles,null,null, false,false);
```
habbes: Add space between comma and argument

src/Templates/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,291 @@
...
+    {
+        if (split) 
+        {
+            EndBlock();
+            string headerText = Template.ToString(_header.Start, _header.Length);
+            string footerText = Template.ToString(_footer.Start, _footer.Length);
+
+            if(_host != null)
+            {
+                Path.GetDirectoryName(_host.TemplateFile);
```
habbes: What's the purpose of calling this method if you are not using the return value?

src/Templates/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,291 @@
...
+    /// <summary>
+    ///Creates a file with the name <paramref name="fileName"> and content <paramref name="content">.
+    /// </summary>
+    /// <param name="fileName">Name of the file to be created</param>
+    /// <param name="content">Content of the file to be created</param>
+    protected virtual void CreateFile(string fileName, string content) 
+    {
+        if (IsFileContentDifferent(fileName, content))
+        {
+                 File.WriteAllText(fileName, content);
```
habbes: Maybe here you should specify which directory it should write to, maybe a temp directory, to avoid it writing at a location that requires privileged access.

habbes: That is assuming this where the error I mentioned occurs from. But I'm not sure if that's the case. If it's not the source of the error, then no need to change it.

paulodero: > Maybe here you should specify which directory it should write to to avoid it writing at a location that requires privileged access.

VS should be able to write to the current directory the class is being executed. will check why you are getting the error.

src/Templates/ODataT4CodeGenFilesManager.ttinclude
paulodero: Should be removed.

src/Templates/ODataT4CodeGenFilesManager.ttinclude
habbes: Could you check what the current directory is? Is it the location where VS is running from or the location where the connected service is loaded from? Or the location where the project that's using the connected service is running from?

src/Templates/ODataT4CodeGenerator.ttinclude
paulodero: Fixed.

src/Templates/ODataT4CodeGenFilesManager.ttinclude
paulodero: removed.

habbes:
@paulodero it seems it seems the file access error has been fixed.

I think now you could just address the case when connected service is being updated then I'll add the approval. i.e.:
- when the user updates the connected service, the "Generate multiple files" checkbox should be set according to the setting saved in the ConnectedService.json file
- This option should be disabled so that the user cannot update the connected service from generating a single service to multiple files or vice versa.

paulodero:
> when the user updates the connected service,

All the cases fixed.

src/Properties/AssemblyInfo.cs
KenitoInc: It's because we don't want to change the original file which had an empty line

paulodero: Reverted. Though I don't think we need the spacing.

PR #32, opened 2020-03-11
Added mocking support.

ElizabethOkerio:


src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// ------------------------------------------------------------------------------
 // <auto-generated>
 //     This code was generated by a tool.
-//     Runtime Version: 16.0.0.0
+//     Runtime Version: 15.0.0.0
```
paulodero: user VS 2019

src/ViewModels/AdvancedSettingsViewModel.cs
```diff
@@ -18,7 +18,7 @@ internal class AdvancedSettingsViewModel : ConnectedServiceWizardPage
         public string GeneratedFileName { get; set; }
         public bool IncludeT4File { get; set; }
         public bool MakeTypesInternal { get; set; }
-        public bool OpenGeneratedFilesInIDE { get; set; } = false;
+        public bool OpenGeneratedFilesInIDE { get; set; }
```
odero: I take it this belongs to a separate PR

ElizabethOkerio: yes. it is an update to the previous merged PR

KenitoInc: Very weird. You should have done a simple fix on a separate PR.

Issue #35, opened 2020-03-12
Add Github templates for contribution and Issues.


marabooy:
Add contribution templates and issue templates for users and update readme with the right sections

PR #36, opened 2020-03-12
Update issue templates

marabooy:
Contributes the first fix to #35 for issue templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+### Describe the bug
+A clear and concise description of what the bug is.
+
+#### Version of the Project affected
```
odero: Would make sense to also include the version of odata client being used

marabooy: This would probably be tied to the OCS release as 0.5.0 will have the 7.6 and 0.4.x had an older version. But let me add as we may have patches that may not be tied to the version

odero: I meant adding that line as one of the things a bug reporter needs to specify since release cycles for OCS and client will defer

PR #38, opened 2020-03-12
Add support for generating VB client proxy/small fixes

unchase:
* Add feature: ~~add option for choosing client proxy generation language (*C#* or *VB*). This option stored in `UserSettings`;~~ client proxy generation language is obtained from target project automatically;
* Add small fix: **AdvancedSettings** link in `AdvancedSettings.xaml` page does not hide the whole `SettingsPanel` *StackPanel*, but only `AdvancedSettingsHyperLinkPanel` *StackPanel*.

VB_generation

* Add small feature: add *VB* and *Web* capabilities to [**ConnectedServiceHandlerExportAttribute.AppliesTo Property**](https://docs.microsoft.com/en-US/visualstudio/extensibility/appliesto-element-visual-studio-templates?view=vs-2019&redirectedfrom=MSDN);
* Add some fixes in code (code refactoring):
  - Use null propagation
  - Inconsistent body style: Use expression-bodied property
  - Convert to auto-property
  - Inline 'out' variable declaration
  - Remove redundant type arguments
  - Use type keyword
  - Delete unused derectives in file
  - Make some fields `readonly`


unchase:
> Is it possible to automatically detect the language of the project and generate the code according to the language detected instead of providing a user with a select option? It would good to reduce the number of input we expect from the users. This will also helps to make the input window less cluttered.

Yes, it is possible. But does the user should to generate the code in the exact language of the project in all cases?

paulodero:
> Is it possible to automatically detect the language of the project and generate the code according to the language detected instead of providing a user with a select option? It would good to reduce the number of input we expect from the users. This will also helps to make the input window less cluttered.
> 
> Yes, it is possible. But does the user should to generate the code in the exact language of the project in all cases?

Yes, that is the most practical use case I can think of.

unchase:
> > Is it possible to automatically detect the language of the project and generate the code according to the language detected instead of providing a user with a select option? It would good to reduce the number of input we expect from the users. This will also helps to make the input window less cluttered.
> > Yes, it is possible. But does the user should to generate the code in the exact language of the project in all cases?
> 
> Yes, that is the most practical use case I can think of.

Done.

test/ODataConnectedService.Tests/CodeGeneration/V4CodeGenDescriptorTest.cs
```diff
@@ -113,6 +113,19 @@ static Project CreateTestProject(string projectPath)
             var projectMock = new Mock<Project>();
             projectMock.SetupGet(p => p.Properties)
                 .Returns(projectPropertiesMock.Object);
+            var projectCodeModelMock = new Mock<CodeModel>();
+            if (targetLanguage == ODataT4CodeGenerator.LanguageOption.CSharp)
+            {
+                projectCodeModelMock.Setup(p => p.Language)
+                    .Returns("{B5E9BD34-6D3E-4B5D-925E-8A43B79820B4}"); // for C# = EnvDTE.CodeModelLanguageConstants.vsCMLanguageCSharp
```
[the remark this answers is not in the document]
unchase: > Why did you prefer to use the literal string here, instead of the constant property?

Due to the following error:

error

unchase:
> Hi @unchase thanks for the contribution. I have tested the feature and it seems to work well. If we're going to add support for VB, however, then we should also bring it up to par with C# support. Here are some feature's that I've noted are not in the generated VB code:
> 
> * Generating methods as virtual to make the mock-able (PR #32)
> * Use of `IDictionary` instead of `Dictionary` as the parameter type for `ByKey()` methods (PR #28)
> 
> Would you mind adding VB support for those features and relevant tests as part of this PR. Thanks.

OK, I will take a look it.

test/ODataConnectedService.Tests/CodeGeneration/V4CodeGenDescriptorTest.cs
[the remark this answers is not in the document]
unchase: I can solve this error by setting `Embed Interop Types` to `False` value:

error2

But I'm not sure it's the right decision.

unchase:
@habbes , done!
Features from PR #28 and PR #32 are supported in code generation for VB now.

test/ODataConnectedService.Tests/CodeGeneration/V4CodeGenDescriptorTest.cs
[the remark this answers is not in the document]
unchase: Apply it?

habbes:
Thanks @unchase, we'll review and get back.

habbes:
@unchase We would like to include this feature in the next release. In order to speed up the review process, and to reduce the chances of conflicts with other open PR's, could you break this PR into two? One for the VB support and the other for refactoring?

habbes:
@unchase actually, don't worry about breaking it up. We'll review the PR soon. Sorry for the confusion.

src/ODataConnectedServiceWizard.cs
```diff
@@ -115,12 +114,13 @@ private ServiceConfiguration CreateServiceConfiguration()
             ServiceConfiguration serviceConfiguration;
             if (ConfigODataEndpointViewModel.EdmxVersion == Common.Constants.EdmxVersion4)
             {
-                var ServiceConfigurationV4 = new ServiceConfigurationV4();
-                ServiceConfigurationV4.IgnoreUnexpectedElementsAndAttributes = AdvancedSettingsViewModel.IgnoreUnexpectedElementsAndAttributes;
-                ServiceConfigurationV4.EnableNamingAlias = AdvancedSettingsViewModel.EnableNamingAlias;
-                ServiceConfigurationV4.IncludeT4File = AdvancedSettingsViewModel.IncludeT4File;
-                ServiceConfigurationV4.OpenGeneratedFilesInIDE = AdvancedSettingsViewModel.OpenGeneratedFilesInIDE;
```
marabooy: `OpenGeneratedFilesInIDE ` Seems to be missing in the other side did you refactor it out?

unchase:
> @unchase actually, don't worry about breaking it up. We'll review the PR soon. Sorry for the confusion.

Ok, thanks.

src/ODataConnectedServiceWizard.cs
unchase: I think taht this option should be available not only for `EdmxVersion4`.

I initialized this variable later in the code and refactored it:
```csharp
        private ServiceConfiguration CreateServiceConfiguration()
        {
            ServiceConfiguration serviceConfiguration;
            if (ConfigODataEndpointViewModel.EdmxVersion == Common.Constants.EdmxVersion4)
            {
                serviceConfiguration = new ServiceConfigurationV4
                {
                    IgnoreUnexpectedElementsAndAttributes =
                        AdvancedSettingsViewModel.IgnoreUnexpectedElementsAndAttributes,
                    EnableNamingAlias = AdvancedSettingsViewModel.EnableNamingAlias,
                    IncludeT4File = AdvancedSettingsViewModel.IncludeT4File
                };
            }
            else
            {
                serviceConfiguration = new ServiceConfiguration();
            }

            serviceConfiguration.ServiceName = ConfigODataEndpointViewModel.ServiceName;
            serviceConfiguration.Endpoint = ConfigODataEndpointViewModel.Endpoint;
            serviceConfiguration.EdmxVersion = ConfigODataEndpointViewModel.EdmxVersion;
            serviceConfiguration.CustomHttpHeaders = ConfigODataEndpointViewModel.CustomHttpHeaders;
            serviceConfiguration.UseDataServiceCollection = AdvancedSettingsViewModel.UseDataServiceCollection;
            serviceConfiguration.GeneratedFileNamePrefix = AdvancedSettingsViewModel.GeneratedFileName;
            serviceConfiguration.UseNameSpacePrefix = AdvancedSettingsViewModel.UseNamespacePrefix;
            serviceConfiguration.MakeTypesInternal = AdvancedSettingsViewModel.MakeTypesInternal;
            serviceConfiguration.OpenGeneratedFilesInIDE = AdvancedSettingsViewModel.OpenGeneratedFilesInIDE;
            serviceConfiguration.GenerateMultipleFiles = AdvancedSettingsViewModel.GenerateMultipleFiles;
            if (AdvancedSettingsViewModel.UseNamespacePrefix && !string.IsNullOrEmpty(AdvancedSettingsViewModel.NamespacePrefix))
            {
                serviceConfiguration.NamespacePrefix = AdvancedSettingsViewModel.NamespacePrefix;
            }

            return serviceConfiguration;
        }
```

marabooy: Thanks for the clarification

unchase:
> Changes look good to me.
> 
> However, there's an error when you opt to generate multiple files. The main file that contains the service context is generated correctly as `Reference.vb` but the remaining files (the ones that contain that entity and enum types) are generated with a `.cs` extension. As a result they are ignored by the VB compiler and the project fails to build. Could you take a look at this? Once fixed the PR should be good to merge.

@habbes fixed

habbes:
@unchase
I think adding vb support when generating multiple files should be a quick fix that would probably only consist in:
- replacing the statements that start a new file such that they don't include extensions in the `ODataT4CodeGenerator.ttinclude`:
```diff
internal void WriteNamespace(string fullNamespace)
{
// ...
-context.MultipleFilesManager.StartNewFile($"{enumType.Name}".cs);
+context.MultipleFilesManager.StartNewFile(enumType.Name);
// ...
-context.MultipleFilesManager.StartNewFile($"{complexType.Name}".cs)
+context.MultpleFilesManager.StartNewFile(complexType.Name);
// ...
-context.MultipleFilesManager.StartNewFile($"{entityType.Name}".cs)
+context.MultipleFilesManager.StartNewFile(entityType.Name)
// ...
}
```
- Adding the appropriate extension based on the selected language option when generating the block `fileName` and `outputFile` in the `FilesManager.GenerateFiles()` method

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1464,7 +1464,14 @@ internal void WriteNamespace(string fullNamespace)
             {
                     if(context.GenerateMultipleFiles) 
                     {
-                        context.MultipleFilesManager.StartNewFile($"{enumType.Name}.cs",false);
+                        if (this.context.TargetLanguage == LanguageOption.VB)
+                        {
+                            context.MultipleFilesManager.StartNewFile($"{enumType.Name}.vb", false);
```
habbes: I think we should pass just the entity name here `context.MultipleFilesManager.StartNewFile(enumType.Name)` and then add the extensions when generating the full filename in the `FilesManager.GenerateFileNames()` method. But this is not blocker, and will approve it as-is.

habbes:
@unchase  thanks for the quick fix, let me test it now and get back to you.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1482,7 +1489,14 @@ internal void WriteNamespace(string fullNamespace)
...
                     {
-                        context.MultipleFilesManager.StartNewFile($"{complexType.Name}.cs",false);
+                        if (this.context.TargetLanguage == LanguageOption.VB)
+                        {
+                            context.MultipleFilesManager.StartNewFile($"{complexType.Name}.vb", false);
+                        }
+                        else
+                        {
+                            context.MultipleFilesManager.StartNewFile($"{complexType.Name}.cs", false);
+                        }
```
paulodero: Could be made shorter probably using ternary operator ( ?: ) to set the extension.

unchase: Done

unchase:
> @unchase Instead of merging changes from master into your branch it is always good to rebase your branch onto OData/master and squash the commits into one commit. This helps us to reduce the conflicts during merge and also create a clean commit history.

I agree with you, this is the right way, I will use it (will rebuild the process) to the following PRs.
Or should the current PR be recycled?

Issue #8, opened 2020-03-03
Allow selecting which actions and functions to generate


habbes:
Related to: https://github.com/OData/lab/issues/15

By default, OData Connected Service generates code for all types, functions and actions defined in the model, many of which may end up unused and polluting the user's application. We should allow the user to select a subset of functions/actions imports to be generated.

unchase:
It would be nice to add the ability to select/exclude operation imports in the UI (in the new Connected Service Wizard page).
Can you implement this in PR #45?

habbes:
@unchase yes definitely, that's the plan. The PR is still a draft precisely because I have not added that part yet. Thanks.

unchase:
In addition to the name of the `OperationImport`, it is desirable to display its signature, since `OperationImports` with the same name can be overloaded.

habbes:
@unchase for this version we're going for a simple approach. When you exclude a name from the list, then all operations with that name regardless of signature (i.e. all overloads) are excluded from the generated code. Since all overloads are excluded, then there's no point in displaying the signature since it doesn't affect anything in this case.

Maybe in a future version we could make it possible to exclude specific overloads of an operation.

PR #45, opened 2020-03-16
Option to exclude some operation imports

habbes:
Addresses issue #8 

This adds an page called "Object Selection" to the wizard where the user can select which action and function imports are included in the generated code:

image

Here are things worth noting about this feature:
- The page only appears if you're connecting to a V4 OData endpoint (this means it will only appear after you've completed the endpoint config on the first page)
- (Update) The listed operations are sorted by name
- When you deselect an item, all action/function imports with that name will be excluded from the generated code regardless of their parameters and return types (i.e. all overloads of the function/action)
- This feature only covers function and action imports and other elements will still be generated as usual (e.g. bound functions/actions which are not declared as operation imports will still be generated)

Initially, I placed the Object Selection page in second position, after the endpoint config because it felt more natural. However I ended up moving it to the end so that it would possible to keep it hidden by default and only dynamically add it when you're leaving the first page. Alternatively, instead of hiding it I could have kept it visible by default and just replace the UI controls with a message saying the feature is only available for V4 endpoints. Let me know if this would have been better.


paulodero:
> Addresses issue #8
> This adds an page called "Object Selection" to the wizard where the user can select which action and function imports are included in the generated code:
> 
> Here are things worth noting about this feature:
> 
> The page only appears if you're connecting to a V4 OData endpoint (this means it will only appear after you've completed the endpoint config on the first page)
> (Update) The listed operations are sorted by name
> When you deselect an item, all action/function imports with that name will be excluded from the generated code regardless of their parameters and return types (i.e. all overloads of the function/action)
> This feature only covers function and action imports and other elements will still be generated as usual (e.g. bound functions/actions which are not declared as operation imports will still be generated)
> 
> Initially, I placed the Object Selection page in second position, after the endpoint config because it felt more natural. However I ended up moving it to the end so that it would possible to keep it hidden by default and only dynamically add it when you're leaving the first page. Alternatively, instead of hiding it I could have kept it visible by default and just replace the UI controls with a message saying the feature is only available for V4 endpoints. Let me know if this would have been better.

@habbes 

1. I think it would be better to make the option available before "setting". We would rather have the tab visible by default. This also reduces the number of actions required from the user incase the user wants to select function imports to add.

2. Why did you decide to name it "Object selection"? Its less intuitive. Why not Select "Function/Action imports"?

habbes:
@paulodero 
In response to your comments:
1. Sure I can make that change, and just disable the page when the endpoint is not v4.
2. I called it Object Selection because I assumed the same page could be used to select entity types as well, not just for functions/actions. If it will only be used for Operation imports, then I can rename it. I got the example for the salesforce connected service pictured in a related issue: https://github.com/OData/lab/issues/15

Should I go ahead and address the changes and update the PR? Asking because you gave an approval.

unchase:
Perhaps should add an `Select/Deselect All` option.
It will be convenient if there are a lot of `OperationImports`.

paulodero:
> Perhaps should add an Select/Deselect All option.
> It will be convenient if there are a lot of OperationImports.

Good idea.

habbes:
@paulodero @unchase 
- I've placed the page after the config endpoint page and now it's always visible. When you connect to a v3 instance, the page's controls are disabled and a message is displaying saying the feature is only available for v4
- I've renamed the page to "Function/Action imports"
- I've added "Select All" and "Unselect All" buttons

habbes:
I also took the liberty of change the text of the "AdvancedSettings" link to "Advanced Settings"

paulodero:
> @paulodero @unchase
> 
> I've placed the page after the config endpoint page and now it's always visible. When you connect to a v3 instance, the page's controls are disabled and a message is displaying saying the feature is only available for v4
> I've renamed the page to "Function/Action imports"
> I've added "Select All" and "Unselect All" buttons

Looks good to me. Please merge.

unchase:
Should rename `ObjectSelectionViewModel_PageEntering` in:

```csharp
OperationImportsViewModel.PageEntering += ObjectSelectionViewModel_PageEntering;
```

PR #46, opened 2020-03-18
Make temp file metadata default

KenitoInc:
We usually store the metadata in a string variable. We have the option of specifying a temp metadata file path where we store the metadata in an xml file. 
This PR makes the temp metadata file the default practice.

### DIFF
--- a/src/CodeGeneration/BaseCodeGenDescriptor.cs
+++ b/src/CodeGeneration/BaseCodeGenDescriptor.cs
@@ -73,16 +73,6 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
             return referenceFolderPath;
         }
 
-        protected string GetEdmxFileFolder()
-        {
-            var serviceReferenceFolderName = this.Context.HandlerHelper.GetServiceArtifactsRootFolder();
-
-            var referenceFolderPath = Path.Combine(
-                ProjectHelper.GetProjectFullPath(this.Project));
-
-            return referenceFolderPath;
-        }
-
         internal async Task CheckAndInstallNuGetPackageAsync(string packageSource, string nugetPackage)
         {
             try
--- a/src/CodeGeneration/BaseCodeGenDescriptor.cs
+++ b/src/CodeGeneration/BaseCodeGenDescriptor.cs
@@ -73,6 +73,16 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
             return referenceFolderPath;
         }
 
+        protected string GetEdmxFileFolder()
+        {
+            var serviceReferenceFolderName = this.Context.HandlerHelper.GetServiceArtifactsRootFolder();
+
+            var referenceFolderPath = Path.Combine(
+                ProjectHelper.GetProjectFullPath(this.Project));
+
+            return referenceFolderPath;
+        }
+
         internal async Task CheckAndInstallNuGetPackageAsync(string packageSource, string nugetPackage)
         {
             try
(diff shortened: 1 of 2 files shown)


src/CodeGeneration/BaseCodeGenDescriptor.cs
```diff
@@ -73,6 +73,16 @@ protected string GetReferenceFileFolder()
...
+        protected string GetEdmxFileFolder()
+        {
+            var serviceReferenceFolderName = this.Context.HandlerHelper.GetServiceArtifactsRootFolder();
+
+            var referenceFolderPath = Path.Combine(
+                ProjectHelper.GetProjectFullPath(this.Project));
+
+            return referenceFolderPath;
+        }
+
```
paulodero: You could reuse `GetReferenceFileFolder()` function instead.

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -104,6 +104,8 @@ private async Task AddGeneratedCSharpCodeAsync()
                 }
             }
             t4CodeGenerator.CustomHttpHeaders = headers;
+            string tempMetadataFile = Path.Combine(GetEdmxFileFolder(), "Edmx" + ".xml");
```
paulodero: You should avoid adding string literals inside code. Use constants or string config files.

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -104,8 +104,11 @@ private async Task AddGeneratedCSharpCodeAsync()
                 }
             }
             t4CodeGenerator.CustomHttpHeaders = headers;
-
             string tempFile = Path.GetTempFileName();
+            const string CsdlFileName = "Csdl.xml";
```
paulodero: I think should be `Edmx.xml` not `Csdl.xml`

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -104,8 +104,11 @@ private async Task AddGeneratedCSharpCodeAsync()
...
+            string tempMetadataFile = Path.Combine(GetReferenceFileFolder(), CsdlFileName);
+            await this.Context.HandlerHelper.AddFileAsync(tempFile, tempMetadataFile);
+            t4CodeGenerator.TempFilePath = tempMetadataFile;
```
paulodero: Better to set a relative path.

KenitoInc: Not necessary since it will always be based on the Reference File Folder

src/CodeGeneration/V4CodeGenDescriptor.cs
KenitoInc: Aren't we standardizing everything from Edmx to Csdl?

src/CodeGeneration/V4CodeGenDescriptor.cs

src/CodeGeneration/BaseCodeGenDescriptor.cs
KenitoInc: #Resolved

src/CodeGeneration/V4CodeGenDescriptor.cs
paulodero: If you move the project to a different directory/path, it wont compile.

gathogojr: @paulodero If you move the project, doesn't the Reference folder not move along? cc. @KenitoInc

KenitoInc: Yes @gathogojr

paulodero: It does, but after generating code you will end up with the following in your generated file for example:

  ```
//more code

private const string filePath = @"C:\Users\paodero\source\repos\CarService\CarService\Connected Services\OData Service\Csdl.xml";

......
                {
                    return global::System.Xml.XmlReader.Create(new global::System.IO.StreamReader(filePath));
                }
```

If you move project the StreamReader will return `FileNotFoundException` unless you generate the code again. This might be a problem if multiple people are working in a project.
We chan change the streamReader to read from the current directory I think.

PR #50, opened 2020-03-19
Add support for loading configs from a config file

ElizabethOkerio:
This PR adds support for loading ODataConnectedService wizard configs from a .json file. 

The following is a sample .json file to load 
<pre>
{
  "ExtendedData": {
    "ExcludedOperationImports": [],
    "EnableNamingAlias": false,
    "IgnoreUnexpectedElementsAndAttributes": false,
    "IncludeT4File": false,
    "ServiceName": "OData Service",
    "Endpoint": "https://services.odata.org/v4/TripPinServiceRW/$metadata",
    "GeneratedFileNamePrefix": "Test",
    "UseNamespacePrefix": true,
    "NamespacePrefix":"Test",
    "UseDataServiceCollection": true,
    "MakeTypesInternal": false,
    "OpenGeneratedFilesInIDE": true,
    "GenerateMultipleFiles":false,
    "CustomHttpHeaders": null,
    "IncludeWebProxy": false,
    "WebProxyHost": null,
    "IncludeWebProxyNetworkCredentials": false,
    "WebProxyNetworkCredentialsUsername": null,
    "WebProxyNetworkCredentialsPassword": null,
    "WebProxyNetworkCredentialsDomain": null,
    "IncludeCustomHeaders": false
  }
}
</pre>
You can as well get this file from an OData Connected Service project and fill the fields as per your requirements. 
Once you fill these fields they will always be pre-loaded whenever you create an OData Connected Service project not unless you want to update them then you can load a new .json file with new configurations.

### DIFF
--- a/src/Common/ExtensionsHelper.cs
+++ b/src/Common/ExtensionsHelper.cs
@@ -1,0 +1,13 @@
+﻿using Microsoft.OData.ConnectedService.Models;
+
+namespace Microsoft.OData.ConnectedService.Common
+{
+    #region For update from Microsoft OData Connected Service
+    internal class ConnectedServiceJsonFileData
+    {
+        public string ProviderId { get; set; }
+
+        public UserSettings ExtendedData { get; set; }
+    }
+    #endregion
+}
--- a/src/ODataConnectedService.csproj
+++ b/src/ODataConnectedService.csproj
@@ -143,6 +143,10 @@
     </Reference>
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\..\RepoConnectedService\ODataConnectedService\packages\Newtonsoft.Json.12.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
     <Reference Include="NuGet.VisualStudio, Version=4.0.0.2323, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\packages\NuGet.VisualStudio.4.0.0\lib\net45\NuGet.VisualStudio.dll</HintPath>
       <EmbedInteropTypes>True</EmbedInteropTypes>
@@ -170,6 +174,7 @@
     <Compile Include="CodeGeneration\V4CodeGenDescriptor.cs" />
     <Compile Include="Common\Constants.cs" />
     <Compile Include="Common\EdmHelper.cs" />
+    <Compile Include="Common\ExtensionsHelper.cs" />
     <Compile Include="Common\ProjectHelper.cs" />
     <Compile Include="Common\UserSettingsPersistenceHelper.cs" />
     <Compile Include="GlobalSuppressions.cs" />
--- a/src/ODataConnectedServiceWizard.cs
+++ b/src/ODataConnectedServiceWizard.cs
@@ -36,7 +36,7 @@ namespace Microsoft.OData.ConnectedService
             this.Context = context;
             this.UserSettings = UserSettings.Load(context.Logger);
 
-            ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.UserSettings);
+            ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.UserSettings, this);
             AdvancedSettingsViewModel = new AdvancedSettingsViewModel(this.UserSettings);
             OperationImportsViewModel = new OperationImportsViewModel();
 
@@ -56,7 +56,7 @@ namespace Microsoft.OData.ConnectedService
 
 
                 // The ViewModel should always be filled otherwise if the wizard is completed without visiting this page the generated code becomes wrong
-                AdvancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNameSpacePrefix;
+                AdvancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNamespacePrefix;
                 AdvancedSettingsViewModel.NamespacePrefix = serviceConfig.NamespacePrefix;
                 AdvancedSettingsViewModel.UseDataServiceCollection = serviceConfig.UseDataServiceCollection;
 
@@ -78,6 +78,7 @@ namespace Microsoft.OData.ConnectedService
                     {
                         var configOdataView = configOdataViewModel.View as ConfigODataEndpoint;
                         configOdataView.Endpoint.IsEnabled = false;
+                        configOdataView.OpenConnectedServiceJsonFileButton.IsEnabled = false;
                         configOdataView.OpenEndpointFileButton.IsEnabled = !serviceConfig.Endpoint.StartsWith("http");
                         configOdataView.ServiceName.IsEnabled = false;
                         configOdataViewModel.IncludeCustomHeaders = serviceConfig.IncludeCustomHeaders;
@@ -101,9 +102,9 @@ namespace Microsoft.OData.ConnectedService
                     {
                         if (advancedSettingsViewModel.View is AdvancedSettings advancedSettings)
                         {
-                            advancedSettingsViewModel.GeneratedFileName = serviceConfig.GeneratedFileNamePrefix;
+                            advancedSettingsViewModel.GeneratedFileNamePrefix = serviceConfig.GeneratedFileNamePrefix;
                             advancedSettings.ReferenceFileName.IsEnabled = false;
-                            advancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNameSpacePrefix;
+                            advancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNamespacePrefix;
                             advancedSettingsViewModel.NamespacePrefix = serviceConfig.NamespacePrefix;
                             advancedSettingsViewModel.UseDataServiceCollection = serviceConfig.UseDataServiceCollection;
                             advancedSettingsViewModel.GenerateMultipleFiles = serviceConfig.GenerateMultipleFiles;
@@ -131,8 +132,7 @@ namespace Microsoft.OData.ConnectedService
         public override Task<ConnectedServiceInstance> GetFinishedServiceInstanceAsync()
         {
             this.UserSettings.Save();
-
-            this.ServiceInstance.InstanceId = AdvancedSettingsViewModel.GeneratedFileName;
+            this.ServiceInstance.InstanceId = AdvancedSettingsViewModel.GeneratedFileNamePrefix;
             this.ServiceInstance.Name = ConfigODataEndpointViewModel.ServiceName;
             this.ServiceInstance.MetadataTempFilePath = ConfigODataEndpointViewModel.MetadataTempPath;
             this.ServiceInstance.ServiceConfig = this.CreateServiceConfiguration();
@@ -161,7 +161,7 @@ namespace Microsoft.OData.ConnectedService
             {
                 serviceConfiguration = new ServiceConfiguration();
             }
-
+            
             serviceConfiguration.ServiceName = ConfigODataEndpointViewModel.ServiceName;
             serviceConfiguration.Endpoint = ConfigODataEndpointViewModel.Endpoint;
             serviceConfiguration.EdmxVersion = ConfigODataEndpointViewModel.EdmxVersion;
@@ -176,8 +176,8 @@ namespace Microsoft.OData.ConnectedService
             serviceConfiguration.IncludeCustomHeaders = ConfigODataEndpointViewModel.IncludeCustomHeaders;
 
             serviceConfiguration.UseDataServiceCollection = AdvancedSettingsViewModel.UseDataServiceCollection;
-            serviceConfiguration.GeneratedFileNamePrefix = AdvancedSettingsViewModel.GeneratedFileName;
-            serviceConfiguration.UseNameSpacePrefix = AdvancedSettingsViewModel.UseNamespacePrefix;
+            serviceConfiguration.GeneratedFileNamePrefix = AdvancedSettingsViewModel.GeneratedFileNamePrefix;
+            serviceConfiguration.UseNamespacePrefix = AdvancedSettingsViewModel.UseNamespacePrefix;
             serviceConfiguration.MakeTypesInternal = AdvancedSettingsViewModel.MakeTypesInternal;
             serviceConfiguration.OpenGeneratedFilesInIDE = AdvancedSettingsViewModel.OpenGeneratedFilesInIDE;
             serviceConfiguration.GenerateMultipleFiles = AdvancedSettingsViewModel.GenerateMultipleFiles;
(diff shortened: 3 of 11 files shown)


ElizabethOkerio:
> Rebase and resolve any conflicts.

Have rebased and resolved conflicts

src/ViewModels/AdvancedSettingsViewModel.cs
```diff
@@ -60,8 +72,6 @@ private void ResetDataContext()
             this.IncludeT4File = false;
             MakeTypesInternal = false;
             this.OpenGeneratedFilesInIDE = false;
-            GenerateMultipleFiles = false;
```
paulodero: Why did you remove this?

unchase:
I think should to use `UserSettings` in `AdvancedSettingsViewModel.cs` to store configuration for each user separately like in [this commit](https://github.com/unchase/Unchase.Odata.Connectedservice/commit/74058cc1ee1d7efa250e1e840ba6e0c06cb872de).
There can be several users on a working machine, each of which generates its own code.

src/Common/ExtensionsHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Microsoft.OData.ConnectedService.Models;
+
+namespace Microsoft.OData.ConnectedService.Common
+{
+    #region For update from Microsoft OData Connected Service
+    internal class ConnectedServiceJsonFileData
+    {
+        public string ProviderId { get; set; }
+
+        public ServiceConfiguration ExtendedData { get; set; }
```
unchase: Why not `UserSettings` instead of `ServiceConfiguration`?

src/ViewModels/AdvancedSettingsViewModel.cs
```diff
@@ -20,6 +21,8 @@ internal class AdvancedSettingsViewModel : ConnectedServiceWizardPage
         public bool MakeTypesInternal { get; set; }
         public bool OpenGeneratedFilesInIDE { get; set; }
         public bool GenerateMultipleFiles { get; set; }
+        public ODataConnectedServiceWizard ServiceWizard { get; set; }
+        public ServiceConfiguration ServiceConfiguration { get; set; }
```
unchase: Why not `UserSettings` instead of `ServiceConfiguration`?

src/ViewModels/AdvancedSettingsViewModel.cs
```diff
@@ -20,6 +21,8 @@ internal class AdvancedSettingsViewModel : ConnectedServiceWizardPage
         public bool MakeTypesInternal { get; set; }
```
unchase: For all properties you can use with `UserSettings`:

```csharp
private bool _makeTypesInternal;
public bool MakeTypesInternal
{
    get => _makeTypesInternal;
    set
    {
        _makeTypesInternal = value;
        UserSettings.MakeTypesInternal = value;
        OnPropertyChanged(nameof(MakeTypesInternal));
    }
}
```

src/ViewModels/AdvancedSettingsViewModel.cs
```diff
@@ -37,6 +40,15 @@ public override async Task OnPageEnteringAsync(WizardEnteringArgs args)
 
             this.View = new AdvancedSettings();
             this.ResetDataContext();
+            if (ServiceConfiguration != null)
+            {
+                this.GeneratedFileName = ServiceConfiguration.GeneratedFileNamePrefix;
```
unchase: For all properties can use with `UserSettings`:
```csharp
this.GeneratedFileName = UserSettings.GeneratedFileName ?? Constants.DefaultReferenceFileName;
```

src/Views/ConfigODataEndpoint.xaml.cs
```diff
@@ -11,9 +17,58 @@ namespace Microsoft.OData.ConnectedService.Views
...
+            if (JObject.Parse(jsonFileText) == null)
+            {
+                MessageBox.Show("Can't convert file content to JObject.", "Open OData Connected Service json-file", MessageBoxButton.OK, MessageBoxImage.Warning);
+                return;
+            }
+
+            var microsoftConnectedServiceData = JsonConvert.DeserializeObject<ConnectedServiceJsonFileData>(jsonFileText);
+            if (microsoftConnectedServiceData != null)
+            {
+                this.ServiceConfiguration = new ServiceConfiguration();
```
unchase: Why not use `UserSettings` instead of `ServiceConfiguration` for all properties?

habbes:
I think it would be a good idea to describe the intention for this feature. I think there is some confusion (at least speaking for myself):

If the intention of the feature is to allow the user to set user-wide default settings that will apply across all future projects and be automatically be loaded to the wizard each time the user adds a new OData Connected Service to a project, then saving them to the User Settings is the way to go. In this scenario, the user would only apply the config file once and will only need to change when they want to set new defaults. Since these settings will apply for clients generated for different services, it's reasonable to expect that only a minimal set of common settings will be set for the user.

However, if your intention was to load the json file each time you're adding a new odata service with those settings, I think using ServiceConfig would make more sense. In this case, the user might have a different json file for different APIs they frequently connect to. Given that you'd have to load the json file each time you want to add a service with preset configuration, I think this would be less useful in practice.

unchase:
> I think it would be a good idea to describe the intention for this feature. I think there is some confusion (at least speaking for myself):
> 
> If the intention of the feature is to allow the user to set user-wide default settings that will apply across all future projects and be automatically be loaded to the wizard each time the user adds a new OData Connected Service to a project, then saving them to the User Settings is the way to go. In this scenario, the user would only apply the config file once and will only need to change when they want to set new defaults. Since these settings will apply for clients generated for different services, it's reasonable to expect that only a minimal set of common settings will be set for the user.
> 
> However, if your intention was to load the json file each time you're adding a new odata service with those settings, I think using ServiceConfig would make more sense. In this case, the user might have a different json file for different APIs they frequently connect to. Given that you'd have to load the json file each time you want to add a service with preset configuration, I think this would be less useful in practice.

Ok, i am agree with you. The better way for this case is using `ServiceConfiguration`.

ElizabethOkerio:
I think it is best to have the wizard automatically loaded each time a user adds a new ODS to a project not unless they want to load different settings. So will use the UserSettings as @unchase suggested

src/ODataConnectedServiceWizard.cs
```diff
@@ -49,7 +49,7 @@ public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
             this.Context = context;
             this.userSettings = UserSettings.Load(context.Logger);
 
-            ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.userSettings);
+            ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.userSettings,this);
```
paulodero: Do you need to pass the whole `ODataConnectedServiceWizard `object in the constructor? If yes, then we do not need to pass `this.userSettings` since it already exist in the object we are passing.

src/ODataConnectedServiceWizard.cs
```diff
@@ -71,9 +67,9 @@ public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
                     {
                         AdvancedSettings advancedSettings = advancedSettingsViewModel.View as AdvancedSettings;
 
-                        advancedSettingsViewModel.GeneratedFileName = serviceConfig.GeneratedFileNamePrefix;
+                        advancedSettingsViewModel.GeneratedFileNamePrefix = serviceConfig.GeneratedFileNamePrefix;
                         advancedSettings.ReferenceFileName.IsEnabled = false;
-                        advancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNameSpacePrefix;
+                        advancedSettingsViewModel.UseNameSpacePrefix = serviceConfig.UseNameSpacePrefix;
```
gathogojr: @ElizabethOkerio Any response for changing this from `UserNamespacePrefix` to `UseNameSpacePrefix`? Namespace is one word...

src/ViewModels/ConfigODataEndpointViewModel.cs
```diff
@@ -1,48 +1,72 @@
...
+        public Version EdmxVersion
+        {
+            get => _edmxVersion;
+            set
+            {
+                _edmxVersion = value;
+                OnPropertyChanged(nameof(EdmxVersion));
+            }
+        }
+        public string MetadataTempPath { get; set; }
```
gathogojr: @ElizabethOkerio Is there a specific reason why we're using automatic getters and setters for some properties and not for others?

ElizabethOkerio: yes. have used automatic getters and setters for values that i'll be loading from UseSettings.

habbes:
Also, when I click the Load from "ConnectedService.json" button and select a file, I don't see the configuration getting loaded in the UI. Maybe I don't know what the fields or format of the file should be. Could you share info about what the expected format/schema of the file is in the description? And does it require the file to be named `ConnectedService.json`?

ElizabethOkerio:
> ConnectedService



> @ElizabethOkerio  Also, when I click the Load from "ConnectedService.json" button and select a file, I don't see the configuration getting loaded in the UI. Maybe I don't know what the fields or format of the file should be. Could you share info about what the expected format/schema of the file is in the description? And does it require the file to be named ConnectedService.json?

So normally when you create proxy classes using OData connected service, it generated a ConnectedService.json file. That is the same structure that needs to be used here. Will document it better.

habbes:
@ElizabethOkerio since the config that is loaded will set new defaults that will be auto-loaded even for connected services that will be added in future projects, I would suggest making this clear to the user. Maybe re-labeling the button to something like "Load new default settings" or a message like "This will set default config that will be auto-loaded whenever you add a new OData Connected Service"

src/ODataConnectedService.csproj
```diff
@@ -143,6 +143,9 @@
     </Reference>
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
     </Reference>
+    <Reference Include="Newtonsoft.Json, Version=12.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
marabooy: Maybe a candidate for the new System.text.Json

marabooy:
Cant wait for you to merge Webproxy to it :)

src/Views/ConfigODataEndpoint.xaml.cs
```diff
@@ -33,5 +39,66 @@ private void OpenMetadataFileButton_OnClick(object sender, RoutedEventArgs e)
...
+            }
+            if (JObject.Parse(jsonFileText) == null)
+            {
+               MessageBox.Show("Can't convert file content to JObject.", "Open OData Connected Service json-file", MessageBoxButton.OK, MessageBoxImage.Warning);
+                return;
+            }
+            var microsoftConnectedServiceData = JsonConvert.DeserializeObject<ConnectedServiceJsonFileData>(jsonFileText);
+            if (microsoftConnectedServiceData != null && this.UserSettings != null)
+            {
+                this.UserSettings.Endpoint = microsoftConnectedServiceData.ExtendedData?.Endpoint ?? this.UserSettings.Endpoint;
```
habbes: Was it necessary to require that the json file have an `ExtendedData` field? Is there any field that's read from the json that's not in the `ExtendedData` object? If not, then why not move the settings to the root object?

src/ViewModels/ConfigODataEndpointViewModel.cs
```diff
@@ -16,61 +16,178 @@ namespace Microsoft.OData.ConnectedService.ViewModels
...
         }
 
         public override async Task OnPageEnteringAsync(WizardEnteringArgs args)
         {
             await base.OnPageEnteringAsync(args);
-            this.View = new ConfigODataEndpoint();
-            this.ResetDataContext();
-            this.View.DataContext = this;
-
             if (PageEntering != null)
```
marabooy: PageEntering?.Invoke

ElizabethOkerio: that's removed part?

src/Views/ConfigODataEndpoint.xaml.cs
```diff
@@ -33,5 +39,66 @@ private void OpenMetadataFileButton_OnClick(object sender, RoutedEventArgs e)
...
             }
         }
-    }
+
+        private void OpenConnectedServiceJsonFileButton_Click(object sender, RoutedEventArgs e)
+        {
+            var openFileDialog = new Microsoft.Win32.OpenFileDialog
+            {
+                DefaultExt = ".json",
+                Filter = "OData Connected Service json-files (.json)|*.json",
```
gathogojr: @ElizabethOkerio (Minor) I'd have gone for filter as `JSON File (.json)|*.json` (or `Json File (.json)|*.json`) and title for this and other dialogs to have the title of `Open OData Connected Service Config File`, i.e., the filter just identifies the file type, while the title just states specifies a config file. If you changed the type of config file to something other than Json you wouldn't need to change the Title

ElizabethOkerio:
1st, 2nd  and 3rd point Fixed. So the loading of settings from a config file is to easen the whole process of having to input them. So even if I load the settings from the config file or I input them they'll be saved to Usersettings. I think that's how it is meant to be. Not that only values loaded from a config file are the ones saved to UseSettings. 
If you load values from a config file..then create a new connected service, the values you loaded from the config file are the ones which will be preloaded. If you loaded values from a config file then made changes to them then when you create a new connected service, the values which will be preloaded are the ones you loaded from a config file and the edited values that you made to the values you loaded.

ElizabethOkerio:
Basically the UserSettings will always be updated depending on the values you input on the wizard. It doesn't matter whether the values are from a config file or they were just entered.

habbes:
> Basically the UserSettings will always be updated depending on the values you input on the wizard. It doesn't matter whether the values are from a config file or they were just entered.

Was that the intended behavior?

So the files does not really set defaults then, maybe I misunderstood the intent of this feature. If that's the case, then you would need to upload that file each time you want to use the settings in the file. And if that's the case, then what's the point of storing the values of the file in User settings?

ElizabethOkerio:
> Basically the UserSettings will always be updated depending on the values you input on the wizard. It doesn't matter whether the values are from a config file or they were just entered.
> 
> Was that the intended behavior?
> So the files does not really set defaults then, maybe I misunderstood the intent of this feature. If that's the case, then you would need to upload that file each time you want to use the settings in the file. And if that's the case, then what's the point of storing the values of the file in User settings?

The values will be stored to UserSettings. No the file does not set the default. The defaults are set by what is saved in UserSettings. The intended behavior was to have each user have their own configs.

habbes:
@ElizabethOkerio 
> The values will be stored to UserSettings. No the file does not set the default. The defaults are set by what is saved in UserSettings. The intended behavior was to have each user have their own configs.

Then I guess I misunderstood the intent of the feature. Having a clear spec on what the feature intended to achieve would have helped.  I did not expect that the values that the user modifies in the UI will become the default.

That said, I think the idea of using the values the user configures for the current connected service as default may or may not be a good idea. For some fields it seems reasonable, but for other fields which are service-specific, it seems unintuitive unless the user is often creating clients for the same service (e.g. service name, service endpoint, custom headers, custom namespace).

ElizabethOkerio:
> @ElizabethOkerio
> 
> The values will be stored to UserSettings. No the file does not set the default. The defaults are set by what is saved in UserSettings. The intended behavior was to have each user have their own configs.
> 
> Then I guess I misunderstood the intent of the feature. Having a clear spec on what the feature intended to achieve would have helped.  I did not expect that the values that the user modifies in the UI will become the default.
> That said, I think the idea of using the values the user configures for the current connected service as default may or may not be a good idea. For some fields it seems reasonable, but for other fields which are service-specific, it seems unintuitive unless the user is often creating clients for the same service (e.g. service name, service endpoint, custom headers, custom namespace).

ok I get your point. will make the changes.

habbes:
@ElizabethOkerio another question regarding the json format, I think some of the fields in that document are not relevant or usable and should not be documented:
- ProviderId
- Version (user's can't set the OCS version)
- GettingStartedDocument (defined by OCS, can't be changed by the user)
- EdmxVersion (version is auto-determined from the document metadata, can't be set by the user)
- ExcludedOperationImports (I think this is service and use-case specific, don't think it should be persisted across connected services)

unchase:
@ElizabethOkerio can you check that issue #69 is resolved by your PR?

Issue #49, opened 2020-03-18
Issues with custom headers feature


habbes:
### Describe the bug
The custom http feature (PR #27) was merged with some bugs and issues that were not caught by the reviewers:
- You cannot add custom http headers when updating the connected service
- An error occurs when the the "Add code templates" feature is enabled
- There should be some help text describing what the feature does, or provides an example of syntax (this is a suggestion, not a bug)
- It would be better if the custom headers feature was hidden by default and toggled if the user wants to see it (similar to how Advanced Settings are toggled) (this is a suggestion, not a bug)

### Version of the Project affected
Issues seem to have been introduced by [this commit](https://github.com/OData/ODataConnectedService/commit/288e7f9b61f8266c4ce5c1dfe421c06d350631d3)

#### 1. Adding custom headers when updating connected service

Create a project and generate client code using the OData Connected Service. Now right-click the connected service from the solution explorer and click "Update Connected Service". You'll notice that the custom headers text box is disabled, if the some headers were necessary for the extension to fetch the metadata from the service, then this means that it becomes effectively impossible for the client code to be updated.

**Proposed solution**: Enable the custom headers field when updating the connected service.

![image](https://user-images.githubusercontent.com/8460169/76973969-aa596f00-6941-11ea-9e42-42e8bb0ad4fe.png)

#### 2. Error when the "Add code templates" setting is enable

When adding an OData Connected Service to your project, if you enable the "Add code templates" option in Advanced settings, it adds the T4 `.tt` and `.ttinclude` template files to your project instead of the generated proxy classes. However, it appears that this feature no longer works correctly since the custom http headers feature was merged.

The following error is thrown when you try to create the service with the option to add code templates enabled.

![image](https://user-images.githubusercontent.com/8460169/76974428-3ec3d180-6942-11ea-96ed-94abf22de56a.png)

It appears that this error only occurs if you leave the custom headers empty. When you add some headers and enable the code templates option, then the error does not occur, and the templates are added as usual.

#### 3. Help text describing what the feature is about

Some people have requested that we add a brief description (or tooltip) explaining what the custom http feature is about (e.g. https://github.com/OData/ODataConnectedService/pull/27#issuecomment-599978324). I think the description should also mention what syntax is expected.

#### 4. Hide the feature by default behind a toggle

[...]

PR #51, opened 2020-03-19
Custom http headers enhancements

KenitoInc:
Fixes issue #49 
This PR fixes regression issues in PR #27 

*Issues fixed*
An exception was being thrown when `custom http headers` was null/empty.
We have enhanced the view by adding a text tool tip for the CustomHttpHeaders textbox
Enabled the ConfigODataEndpoint view during Connected service update

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -72,13 +72,14 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
                 text = Regex.Replace(text, "(public const bool EnableNamingAlias = )true;", "$1" + this.ServiceConfiguration.EnableNamingAlias.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
                 text = Regex.Replace(text, "(public const bool IgnoreUnexpectedElementsAndAttributes = )true;", "$1" + this.ServiceConfiguration.IgnoreUnexpectedElementsAndAttributes.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
                 text = Regex.Replace(text, "(public const bool MakeTypesInternal = )false;", "$1" + ServiceConfiguration.MakeTypesInternal.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
-                text = Regex.Replace(text, "(public const bool CustomHttpHeaders = )\"\";", "$1" + ServiceConfiguration.CustomHttpHeaders.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
+                text = Regex.Replace(text, "(public const string CustomHttpHeaders = )\"\";", "$1@\"" + ServiceConfiguration.CustomHttpHeaders ?? string.Empty + "\";");
                 await writer.WriteAsync(text);
                 await writer.FlushAsync();
             }
 
             string referenceFolder = GetReferenceFileFolder();
             await this.Context.HandlerHelper.AddFileAsync(Path.Combine(t4Folder, "ODataT4CodeGenerator.ttinclude"), Path.Combine(referenceFolder, this.GeneratedFileNamePrefix + ".ttinclude"));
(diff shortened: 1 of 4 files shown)


src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -72,7 +72,12 @@ private async Task AddT4FileAsync()
                 text = Regex.Replace(text, "(public const bool EnableNamingAlias = )true;", "$1" + this.ServiceConfiguration.EnableNamingAlias.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
                 text = Regex.Replace(text, "(public const bool IgnoreUnexpectedElementsAndAttributes = )true;", "$1" + this.ServiceConfiguration.IgnoreUnexpectedElementsAndAttributes.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
                 text = Regex.Replace(text, "(public const bool MakeTypesInternal = )false;", "$1" + ServiceConfiguration.MakeTypesInternal.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
-                text = Regex.Replace(text, "(public const bool CustomHttpHeaders = )\"\";", "$1" + ServiceConfiguration.CustomHttpHeaders.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
+                string CustomHeaders = "";
```
unchase: Better:

```csharp
string customHeaders = ServiceConfiguration.CustomHttpHeaders ?? string.Empty;
text = Regex.Replace(text, "(public const string CustomHttpHeaders = )\"\";", "$1@\"" + customHeaders + "\";");
```

or 

```csharp
text = Regex.Replace(text, "(public const string CustomHttpHeaders = )\"\";", "$1@\"" + ServiceConfiguration.CustomHttpHeaders ?? string.Empty + "\";");
```

marabooy:
@KenitoInc could you also split the headers using both LF and CRLF. the user could have different settings or copy paste them from a source that is not the Environment.NewLine

src/CodeGeneration/V4CodeGenDescriptor.cs
KenitoInc: Fixed.

src/Views/ConfigODataEndpoint.xaml
```diff
@@ -38,6 +38,14 @@
             AcceptsReturn="True"
             VerticalScrollBarVisibility="Visible"
             Margin="20, 5, 10, 5"
-            Width="250"/>
+            Width="250">
+            <TextBox.ToolTip>
+                <ToolTip HorizontalAlignment="Center">
+                    <TextBlock>
+                        (Optional) We add Http Headers to our http request as a multiline string
```
habbes: Could you show an example of the actual syntax, saying "multiline string" alone is not clear? e.g.:

```
Header1: value1
Header2: value2
```

KenitoInc:
> The null reference error when you opt to add templates is no longer there.
> However, the user is still unable to set headers when updating the connected service.

We need to make a decision if we need to enable updating all fields in ConfigODataEndpoint view, not just CustomHttpHeaders

habbes:
@KenitoInc 

> We need to make a decision if we need to enable updating all fields in ConfigODataEndpoint view, not just CustomHttpHeaders

The motivation for allowing the user to edit the custom headers field when updating the connected service is because custom headers are not saved in the `ConnectedService.json` file (for security reasons). But all the other fields are stored. So if you're creating a connected service for an endpoint that requires certain headers (e.g. auth headers) in order to return the metadata document, then you will also need to set those same headers manually when updating the connected service, otherwise you will not be able to retrieve the metadata and therefore will be unable to update the connected service.

ElizabethOkerio:
Please a good description of what the PR does.

src/Views/ConfigODataEndpoint.xaml
KenitoInc: #Resolved

PR #53, opened 2020-03-20
Increment version for dev

habbes:
Now that 0.6.0 has been release, let's bump the development version so that it doesn't conflict with the version in production in case we have both installed.


habbes:
Closed cause the branch wasn't properly rebased.

PR #66, opened 2020-03-25
Using var instead of types

### LINKED ISSUE #42: Using var instead of types
Should replace specific types with the `var` keyword  in many places in the code.

For example:

var
...and many others

unchase:


unchase:
@habbes , @ElizabethOkerio or @paulodero can you set **ready for review** label?

gathogojr:
To var or not to var, that is the question...

@ElizabethOkerio Personally, I'm not that attracted to the blanket use of `var`. I find that it makes code harder to read/understand in specific scenarios. E.g.,
```
var result = DoSomething();
```
It makes easier reading if I can tell what the return type of the `DoSomething()` method is by just looking at the statement,
```
Type result = DoSomething();
```
This is one example of `var` usage that I feel unattracted to. Which is why generally in projects with multiple contributors I'd prefer specifying the variable type explicitly. I don't mind much when var is used in initialization statements like these:
```
var a = new Type();
```
All the same, I'd be eager to hear other people's opinion on this

paulodero:
> To var or not to var, that is the question...
> @ElizabethOkerio @unchase Personally, I'm not that attracted to the blanket use of var. I find that it makes code harder to read/understand in specific scenarios. E.g.,
> var result = DoSomething();
> 
> It makes easier reading if I can tell what the return type of the DoSomething() method is by just looking at the statement,
> Type result = DoSomething();
> 
> This is one example of var usage that I feel unattracted to. Which is why generally in projects with multiple contributors I'd prefer specifying the variable type explicitly. I don't mind much when var is used in initialization statements like these:
> var a = new Type();
> 
> All the same, I'd be eager to hear other people's opinion on this

@gathogojr I agree. According to [C# coding convention](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions) you only use `var `when the the type of the variable is obvious from the assignment as you have explained. In this case some variables are while others are not. We can use this rule to choose which ones to use `var`.

KenitoInc:
@g2mula Join this discussion

src/CodeGeneration/V3CodeGenDescriptor.cs
```diff
@@ -35,31 +35,31 @@ public override async Task AddGeneratedClientCodeAsync()
...
-            using (XmlReader reader = XmlReader.Create(this.MetadataUri, settings))
+            using (var reader = XmlReader.Create(this.MetadataUri, settings))
             {
-                string tempFile = Path.GetTempFileName();
-                bool noErrors = true;
+                var tempFile = Path.GetTempFileName();
+                var noErrors = true;
 
-                using (StreamWriter writer = File.CreateText(tempFile))
+                using (var writer = File.CreateText(tempFile))
```
habbes, line 62: I would use type here, might not be directly obvious that these returns a `StreamWriter`

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -119,7 +119,7 @@ private async Task AddGeneratedCodeAsync()
         {
             await this.Context.Logger.WriteMessageAsync(LoggerMessageCategory.Information, "Generating Client Proxy for OData V4...");
 
-            ODataT4CodeGenerator t4CodeGenerator = CodeGeneratorFactory.Create();
+            var t4CodeGenerator = CodeGeneratorFactory.Create();
```
habbes, line 122: Not obvious what the type of `t4CodeGenerator` is to someone reading the code.

src/Common/ProjectHelper.cs
```diff
@@ -20,7 +20,7 @@ internal static class ProjectHelper
         public static Project GetProjectFromHierarchy(IVsHierarchy projectHierarchy)
         {
             object projectObject;
-            int result = projectHierarchy.GetProperty(
+            var result = projectHierarchy.GetProperty(
```
habbes, line 23: Not obvious that `result` is an int

src/Common/UserSettingsPersistenceHelper.cs
```diff
@@ -20,23 +20,23 @@ internal class UserSettingsPersistenceHelper
...
         public static void Save(object userSettings, string providerId, string name, Action onSaved, ConnectedServiceLogger logger)
         {
-            string fileName = UserSettingsPersistenceHelper.GetStorageFileName(providerId, name);
+            var fileName = UserSettingsPersistenceHelper.GetStorageFileName(providerId, name);
 
             UserSettingsPersistenceHelper.ExecuteNoncriticalOperation(
                 () =>
                 {
-                    using (IsolatedStorageFile file = UserSettingsPersistenceHelper.GetIsolatedStorageFile())
+                    using (var file = UserSettingsPersistenceHelper.GetIsolatedStorageFile())
```
habbes, line 28: Not obvious what type `file` is

src/ODataConnectedServiceHandler.cs
```diff
@@ -51,7 +51,7 @@ private async Task<BaseCodeGenDescriptor> SaveServiceInstanceAsync(ConnectedServ
 
         private async Task<BaseCodeGenDescriptor> GenerateCodeAsync(string metadataUri, Version edmxVersion, ConnectedServiceHandlerContext context, Project project)
         {
-            BaseCodeGenDescriptor codeGenDescriptor = codeGenDescriptorFactory.Create(edmxVersion, metadataUri, context, project);
+            var codeGenDescriptor = codeGenDescriptorFactory.Create(edmxVersion, metadataUri, context, project);
```
habbes, line 54: Not obvious what type `codeGenDescriptor` is

unchase:
> I also agree that we should not use `var` indiscriminately. I have highlighted a few examples in the PR where the type of the variable were not obvious to me. We would also not use `var` when "down-casting" to a base type or interface, e.g. `IEdmEntityType entityType = new EdmEntityType`

@habbes I agree with your comments. Fixed it.

KenitoInc:
@unchase Rebase, resolved conflicts and also check if there are more variables that need to be updated

unchase:
> @unchase Rebase, resolved conflicts and also check if there are more variables that need to be updated

@KenitoInc done!

Issue #44, opened 2020-03-13
Allow to set labels for created issues and PRs


unchase:
Is it possible to give the opportunity to set Labels for own created issues and PRs?

What types of Labels are generally available?

marabooy:
Seems like gihub doesn't support this as the member should have write access to the repo for it to work. https://help.github.com/en/github/managing-your-work-on-github/applying-labels-to-issues-and-pull-requests

unchase:
Possible solutions:

* https://github.com/ZELLMECHANIK-DRESDEN/ShapeOut/issues/258#issuecomment-551516384
* [add more **ISSUE_TEMPLATE**](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates) with various labels

@habbes could you watch it?

habbes:
It seems this has now been sorted @unchase ?

unchase:
> It seems this has now been sorted

@habbes How?

habbes:
@unchase I've seen that you have been able to add labels to issues.

unchase:
> @unchase I've seen that you have been able to add labels to issues.

@habbes Unfortunately I can not

unchase:
Can you resolve the issue?
This would greatly simplify the work with PRs and issues.

PR #68, opened 2020-03-25
Add ability for choosing local metadata files

unchase:
browse

### DIFF
--- a/src/ODataConnectedServiceWizard.cs
+++ b/src/ODataConnectedServiceWizard.cs
@@ -78,6 +78,7 @@ namespace Microsoft.OData.ConnectedService
                     {
                         var configOdataView = configOdataViewModel.View as ConfigODataEndpoint;
                         configOdataView.Endpoint.IsEnabled = false;
+                        configOdataView.OpenEndpointFileButton.IsEnabled = !serviceConfig.Endpoint.StartsWith("http");
                         configOdataView.ServiceName.IsEnabled = false;
                         configOdataViewModel.IncludeCustomHeaders = serviceConfig.IncludeCustomHeaders;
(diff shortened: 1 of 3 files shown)


habbes:
Hi @unchase, thanks for this contribution. Could you rebase your branch.

unchase:
> Hi @unchase, thanks for this contribution. Could you rebase your branch.

@habbes done.

marabooy:
@unchase  could you mark the branch as ready for review if the branch is ready. Is there any pending work?

unchase:
> @unchase could you mark the branch as ready for review if the branch is ready. Is there any pending work?

@marabooy Tell me, how can I do this ?

labels

marabooy:
Is the cog on the right accessible? If not just mention @habbes , @ElizabethOkerio  or @paulodero and they will do it for you.
image

marabooy:
> @unchase The button allows you to change the service endpoint even when the text field is disabled. I think the button should be disabled when the text field is disabled.
> 
> Everything else looks fine to me.
@habbes maybe for files we wan't to leave it that way? as they may have moved the file or deleted it. unless if we want to copy the file to the project. It may be hard to enforce the same file location for local files. 
@unchase  I think what we can do (in my opinion) is if the file is local leave it as is but if it was a url disable the button.

habbes:
@marabooy @unchase good point, being able to update the file sounds a like a good idea.

unchase:
> Is the cog on the right accessible? If not just mention @habbes , @ElizabethOkerio or @paulodero and they will do it for you.
> image

Why not give contributors such an opportunity? (issue #44)

unchase:
@habbes @marabooy 
Done.
web

file

PR #71, opened 2020-03-26
Select Schema types

paulodero:
Allow users to select entity types to generate.


habbes:
@paulodero could you share the behavior that should be expected to occur in the following scenarios:
- you exclude an entity but it's been referenced (either as a relationship/navigation property or function parameter/return type) of an entity that's been included
- you exclude an entity but it's been referenced as a parameter or return type of an included function/action import

paulodero:
> @paulodero could you share the behavior that should be expected to occur in the following scenarios:
> 
> you exclude an entity but it's been referenced (either as a relationship/navigation property or function parameter/return type) of an entity that's been included
> you exclude an entity but it's been referenced as a parameter or return type of an included function/action import

@habbes incase any of the above mentioned relationships exist, the expectation is that if you select one type, the related types are also selected automatically from the UI.

habbes:
@paulodero please update your branch, thanks.

src/ViewModels/OperationImportsViewModel.cs
```diff
@@ -80,11 +80,32 @@ public void LoadOperationImports(IEnumerable<IEdmOperationImport> operationImpor
...
+
+                    operationImportModel.PropertyChanged += (s, args) =>
+                    {
+                        IEnumerable<IEdmOperationParameter> parameters = operation.Operation.Parameters;
+
+                        foreach (var parameter in parameters)
+                        {
+                            if (schemaTypeModels.TryGetValue(parameter.Type.FullName(), out SchemaTypeModel model) && !model.IsSelected)
+                            {
+                                model.IsSelected = true;
```
habbes, line 97: What if `opertionImportModel` is getting unselected?

src/ViewModels/OperationImportsViewModel.cs
```diff
@@ -80,11 +80,32 @@ public void LoadOperationImports(IEnumerable<IEdmOperationImport> operationImpor
...
                     {
                         Name = operation.Name,
-                        IsSelected = true
-                    });
+                        IsSelected = IsOperationImportIncluded(operation, excludedSchemaTypes)
+                    };
+
+                    operationImportModel.PropertyChanged += (s, args) =>
+                    {
+                        IEnumerable<IEdmOperationParameter> parameters = operation.Operation.Parameters;
```
habbes, line 91: I think instead of defining this delegate here and modifying the schema type models from inside the `LoadOperationImports`, it would be better to pass a delegate or Action to `LoadOperationImports`, and that Action, passed as the event handler to `operationImportModel.PropertyChanged += operationImportChangedHandler`. And then when you could add the actual `Operation` as a property of `OperationImportModel` and get the parameters from there, that way the event handler would not need to be a closure.

I think this would make things cleaner and more decoupled, and probably better for performance cause you would need only one event handler action, instead of each operation model having its own event handler instance.

src/ViewModels/SchemaTypesViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+                _isSupportedVersion = value;
+                OnPropertyChanged("VersionWarningVisibility");
+            }
+        }
+
+        public Visibility VersionWarningVisibility
+        {
+            get
+            {
+                return IsSupportedODataVersion ? Visibility.Hidden : Visibility.Visible;
```
habbes, line 52: I think the built-in BooleanToVisibilityConverter could be used for this.

src/Common/EdmHelper.cs
```diff
@@ -52,5 +62,56 @@ public static IEnumerable<IEdmOperationImport> GetOperationImports(IEdmModel mod
                 }
             }
         }
+
+        public static IDictionary<IEdmStructuredType,List<IEdmOperation>> GetBoundOperations(IEdmModel model)
+        {
+            if(_boundOperations != null)
+            {
+                return _boundOperations;
```
habbes, line 70: Is it possible that the `IEdmModel` passed here would be different from the one passed in the previous call to this method? If that's the case, won't the method return the wrong operations?

src/ViewModels/OperationImportsViewModel.cs
paulodero: This should not impact the selection of the `SchemaTypes ` expecially if the user explicitly selected the specific type. Remember the types are not only used by operation imports. We automatically select Schematypes to avoid chances of the user generating uncompilable code.

src/ViewModels/OperationImportsViewModel.cs
paulodero: Good idea. Although each model is treated as a separate object and would still eventually has its own handler.

src/Common/EdmHelper.cs
paulodero: I don't think that is possible while using connected service. You only pass one `$metadata` file which should have one `IEdmModel`. This ensures the model is not parsed everytime this function is called.

src/ViewModels/SchemaTypesViewModel.cs
paulodero: Actually this is not even required. This feature is applicable to all OData versions. Refactoring this.

ElizabethOkerio:
@KenitoInc  you've summarized the issues really well. Also had them. Although if I select person all the related entities are selected. But I do not want to get all the related entities to person. I should be able to unselect so I remain with the ones I only want.

KenitoInc:
> @KenitoInc  you've summarized the issues really well. Also had them. Although if I select person all the related entities are selected. But I do not want to get all the related entities to person. I should be able to unselect so I remain with the ones I only want.

Makes sense

habbes:
> @KenitoInc  you've summarized the issues really well. Also had them. Although if I select person all the related entities are selected. But I do not want to get all the related entities to person. I should be able to unselect so I remain with the ones I only want.

I think being able to unselect the ones you don't want, even if the generated code fails to compile is exactly what the feature wanted to prevent.

I think the way he has made it work with operation imports is what was intended. If you select an operation import that references an entity as a return or parameter type, then that entity is automatically selected. If you go back and unselect that entity, then the operation import gets unselected as well.

paulodero:
> @paulodero @habbes
> The feature works relatively fine, however there are a few loose ends we may think about as improvements:
> Scenario 1: Location (Complex Type) has City (Complex type) as a property. When I select Location, City is automatically selected. The converse is not true. This behaviour is okay but when we only select City, we will have an “orphan” complex type which has not been exposed via the entity container. The user cannot use the Complex Type. #Edge Case
> Scenario 2: In some instances when you select on entity type(and the related schema types), the generated code doesn’t compile since some extension methods generate depend on other entities not generated. E.g Select Airport( and AirportLocation) #Bug
> Scenario 3: When I select Person, the related entities are autoselected. However I can unselect them. We can thing of how we can enhance this in a future release. #Enhancement

@KenitoInc For scenario 1 and scenario 3, we may not focus on that since to do the reverse check you need to search through the whole supposedly to be generated code which for performance purposes is too much work for the program while adding little value and also denying the user flexibility of generating types he wants. For automatic selection, the aim is to ensure the user does not generate uncompilable code. 

Let me take a look on scenario 2.

habbes:
@paulodero 

> @KenitoInc For scenario 1 and scenario 3, we may not focus on that since to do the reverse check you need to search through the whole supposedly to be generated code which for performance purposes is too much work for the program while adding little value and also denying the user flexibility of generating types he wants. For automatic selection, the aim is to ensure the user does not generate uncompilable code.
> Let me take a look on scenario 2.

But for scenario 3, would it not lead to uncompilable code? If you unselect something that was auto-selected, then it means you're unselected a required dependency and your code will fail to compile. Or am I getting it wrong?

paulodero:
> @paulodero
> If you change the service endpoint before completing the config, the feature still displays the entities of the first endpoint.
> e.g.:
> 
> set the service endpoint to something like trippin
> click next to include/exclude entities and operation imports
> go back to first page and change the endpoint
> click next, the entities and operation imports displayed are still from the first endpoint. One would expect it to update the list with the types and operations from the updated endpoint

Got it this is where the singleton bring the problem. Will fix in a few minutes.

habbes:
@paulodero, again regarding performance. What about this: if someone unselects an entity, then you iterate over the currently selected entities and reselect them, this will ensure that their dependencies are also automatically reselected. And if the unselected one was a dependency, then it's automatically reselected as well, and so we wouldn't have the compilation error. Would this still be as costly as a reverse search performance-wise?

I guess in the worst case, if you have everything selected, then unselect one, it would be just as bad as a reverse check since you would have to go through all the entities in the model but one. But if you cache the dependencies for each entity, then maybe it's not that bad performance-wise. If it's not overkill, I think this could help ensure the user does not unselect something that's required for compilation to succeed.

paulodero:
> @paulodero
> 
> @KenitoInc For scenario 1 and scenario 3, we may not focus on that since to do the reverse check you need to search through the whole supposedly to be generated code which for performance purposes is too much work for the program while adding little value and also denying the user flexibility of generating types he wants. For automatic selection, the aim is to ensure the user does not generate uncompilable code.
> Let me take a look on scenario 2.
> 
> But for scenario 3, would it not lead to uncompilable code? If you unselect something that was auto-selected, then it means you're unselected a required dependency and your code will fail to compile. Or am I getting it wrong?
> I think the performance argument make sense, but for flexibility, I don't think you can offer the user flexibility of choosing types they want and still guarantee the code is compilable. I think it's okay to compromise flexibility in order to guarantee correctly generated code.

Correct! By autoselecting I think its a guidance enough to the user. I don't thing we need to slow down the selection process for this sake.

paulodero:
> @paulodero, again regarding performance. What about this: if someone unselects an entity, then you iterate over the currently selected entities and reselect them, this will ensure that their dependencies are also automatically reselected. And if the unselected one was a dependency, then it's automatically reselected as well, and so we wouldn't have the compilation error. Would this still be as costly as a reverse search performance-wise?
> I guess in the worst case, if you have everything selected, then unselect one, it would be just as bad as a reverse check since you would have to go through all the entities in the model but one. But if you cache the dependencies for each entity, then maybe it's not that bad performance-wise. If it's not overkill, I think this could help ensure the user does not unselect something that's required for compilation to succeed.

The default is that all types are selected. To implement this while unselecting means you have to go throught all the types/bound and unbvound functions and actions/return types and params. I don't think its worth the effort. If it is it can be made available in the next iteration.

marabooy:
@paulodero For usability you may need a new panel/ui to showcase which entities will be excluded or a way to notify users that some unselected entities will still be included due to their dependencies. 

Question:
If one opts to remove a certain navigational property can you generate the code without that property, and what would it mean to remove it?

habbes:
@marabooy 
> Question:
> If one opts to remove a certain navigational property can you generate the code without that property, and what would it mean to remove it?

If I recall correctly, it was decide that entities would always be generated in their entirity. You either generate the whole entity, or don't generate it at all. But you can't arbitrarily pick which properties of an entity to include/exclude.

marabooy:
@habbes but then not generating say Authors which has a navigation to say Books which is being generated Implies that Authors will  be Included. Which is something we need to show to the user to make sure its intuitive to them. And an easy way would be to have a heading stating that navigation elements will be added automatically then automatically selecting them when the "parent" is selected

habbes:
@marabooy I think we could add a note along the lines  "Types which are used by the selected types will also be generated". I believe that would be easier than creating a new panel and still helps the user understand why some types are autoselected when you select others.

habbes:
@paulodero 
> The default is that all types are selected.

I think this is why generation of uncompilable code is more likely to occur. The fact that all types are selected by default (which definitely a good thing) means that if the user wants to remove a few types, then they will start unselecting. But if they unselect, they might unselect types that are required by the selected types without knowing. The workaround for this would be for the user to unselect all types first, then start selecting the ones they need. Because of this workaround, maybe it's something we can leave out from this iteration, but definitely something I think we should consider again down the line.

habbes:
@paulodero I think the feature does not account for inherited types. i.e. I tested an endpoint with `AttendeeBase` which is a base class, and `Attendee` as subtype. Assuming all are unselected, if I select `Attendee`, then naturally it should auto-select `AttendeeBase` because `Attendee` depends on `AttendeeBase`. But it doesn't auto-select it.

paulodero:
> @paulodero I think the feature does not account for inherited types. i.e. I tested an endpoint with AttendeeBase which is a base class, and Attendee as subtype. Assuming all are unselected, if I select Attendee, then naturally it should auto-select AttendeeBase because Attendee depends on AttendeeBase. But it doesn't auto-select it.

Yes, as I mentioned, it doesn't. Will be taken care of in the next iteration.

paulodero:
> @paulodero
> 
> The default is that all types are selected.
> 
> I think this is why generation of uncompilable code is more likely to occur. The fact that all types are selected by default (which definitely a good thing) means that if the user wants to remove a few types, then they will start unselecting. But if they unselect, they might unselect types that are required by the selected types without knowing. The workaround for this would be for the user to unselect all types first, then start selecting the ones they need. Because of this workaround, maybe it's something we can leave out from this iteration, but definitely something I think we should consider again down the line.

Correct. Will be available in the next iteration.

paulodero:
> @marabooy
> 
> Question:
> If one opts to remove a certain navigational property can you generate the code without that property, and what would it mean to remove it?
> 
> If I recall correctly, it was decide that entities would always be generated in their entirity. You either generate the whole entity, or don't generate it at all. But you can't arbitrarily pick which properties of an entity to include/exclude.

Correct. UI improvements can be done in the next iteration if need be.

paulodero:
> @paulodero For usability you may need a new panel/ui to showcase which entities will be excluded or a way to notify users that some unselected entities will still be included due to their dependencies.
> Question:
> If one opts to remove a certain navigational property can you generate the code without that property, and what would it mean to remove it?

Good suggestions. Could we have these UI improvements in our subsequent iterations?

habbes:
@paulodero  @marabooy  @KenitoInc 
Perhaps this is a scenario where it's useful to allow the user to unselects types even if it it's needed by other types: https://github.com/OData/ODataConnectedService/issues/81 in that scenario the user has already implemented that type and prefers that the code generator skips it.

paulodero:
> @paulodero  @marabooy  @KenitoInc
> Perhaps this is a scenario where it's useful to allow the user to unselects types even if it it's needed by other types: #81 in that scenario the user has already implemented that type and prefers that the code generator skips it.

Sure, let's see the use's response to the released feature and make improvements where there is need.

src/ViewModels/SchemaTypesViewModel.cs
paulodero: fixed

src/Common/EdmHelper.cs
paulodero: The function is now able to detect if the model changed and create a new model.

weitzhandler:
Fixes #81.
I doubt if [PR Linking](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) would work here.

paulodero:
> @paulodero
> If you change the service endpoint before completing the config, the feature still displays the entities of the first endpoint.
> e.g.:
> 
> set the service endpoint to something like trippin
> click next to include/exclude entities and operation imports
> go back to first page and change the endpoint
> click next, the entities and operation imports displayed are still from the first endpoint. One would expect it to update the list with the types and operations from the updated endpoint

Fixed.

habbes:
@paulodero found that in some cases the feature generates code that does not compile even when you explicitly select types you need (which should auto-select dependencies).

In this particular example using Trippin, when you unselect all then select "Person", it also selects related types like PersonGender. However the code does not compile because AirlineSingle is referenced, but it was not defined. I tried to update the connected service to include the type, it was not obvious which type to select because AirlineSingle itself is not on the list, so I selected Airline. However, I got the same error that AirlineSingle is referenced but not defined. It turns out that you needed to select Airport in order for AirlineSingle to be generated. This was definitely not obvious.

Another curious behavior is that if you "Select All" types, then unselect just the Airport, when you generate code and check connectedservice.json, you realize that Airport was not actual excluded. It's as if there's something else that causes it to re-auto select. Not sure what.

paulodero:
> @paulodero found that in some cases the feature generates code that does not compile even when you explicitly select types you need (which should auto-select dependencies).
> In this particular example using Trippin, when you unselect all then select "Person", it also selects related types like PersonGender. However the code does not compile because AirlineSingle is referenced, but it was not defined. I tried to update the connected service to include the type, it was not obvious which type to select because AirlineSingle itself is not on the list, so I selected Airline. However, I got the same error that AirlineSingle is referenced but not defined. It turns out that you needed to select Airport in order for AirlineSingle to be generated. This was definitely not obvious.
> Another curious behavior is that if you "Select All" types, then unselect just the Airport, when you generate code and check connectedservice.json, you realize that Airport was not actual excluded. It's as if there's something else that causes it to re-auto select. Not sure what.

Same issue reported by @KenitoInc  and is being worked on.

habbes:
@paulodero that issue seems fixed on my end. I still occasionally get uncompilable code because the auto-selection of dependencies does not detect required based types. But I believe you said that is the subject of a future iteration.

PR #73, opened 2020-03-26
Bump version to 0.7.0

habbes:


src/ODataConnectedServiceProvider.cs
```diff
@@ -20,7 +20,7 @@ public ODataConnectedServiceProvider()
             Description = "OData Connected Service for V1-V4";
             Icon = new BitmapImage(new Uri("pack://application:,,/" + this.GetType().Assembly.ToString() + ";component/Resources/Icon.png"));
             CreatedBy = "OData";
-            Version = new Version(0, 6, 1);
+            Version = new Version(0, 7, 0);
```
unchase: May be should use:
```csharp
Version = typeof(ODataConnectedServiceProvider).Assembly.GetName().Version;
```
To not change 3 times.

PR #78, opened 2020-03-27
Fix issue #69: saving the wizard pages state

### LINKED ISSUE #69: Save the wizard pages state
It would be nice to save the state of the wizard pages when a user navigates from page to page.

For example, the user selected the `OperationImports` to add, switched to another page, and then decided to change the selected `OperationImports`. 
With current behavior, he will have to choose everything anew.

Demostration:
save_state

### LINKED ISSUE #94: Name of last connected service is used when service name is left as default
### Describe the bug
When you create a connected service without modifying the service name, i.e. leaving the default "OData Service", the generated service is saved under the name of a previously generated connected service.

ODataConnectedService on master branch (as of commit [46f7866](https://github.com/OData/ODataConnectedService/commit/46f78667e31cc8629774d4cd91aeaf7f8a2d7b7b))

#### To Reproduce
1. Create connected service, give it a service name, e.g. "ServiceA" and complete the code generation process
2.  Create another connected service, leave the default service name "OData Service" unchanged. Complete the code generation process.

#### Expected behavior
The code and files for the second connected service should be saved under "OData Service" subfolder with "OData Service" saved a the service name

#### Actual behavior
The generated code is saved under "ServiceA" and overwrites the code and files in the "ServiceA" subfolder. Furthermore the "ServiceName" in the generated ConnectedService.json file is saved as "ServiceA".

#### Additional context
- If you modify the service name field before code generation, the bug does not occur
- If you load a config file including the service name, the bug does not occur
- In the example provided in the repro-steps, if you were to create another service in between the two that loads the service name and other settings from a config file, the third service with a default service name would still be saved under "ServiceA", it seems that it picks the service name of the last saved connected service that did not load settings from a config file. Maybe the service name is saved in user settings when editing the field manually, but not saved in user settings when loading from a config file
- The bug did not occur in v0.7.1 that's on Marketplace, it seems to be a regression introduced by PR #50

unchase:
Fix #69, fix #94

Refactored `ODataConnectedServiceWizard`:
* extract `PageEntering`-methods
* rename `ObjectSelectionViewModel_PageEntering` to `OperationImportsViewModel_PageEntering`
* update **Operation Imports** only for the new Endpoints

save_state_new

### DIFF
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -18,7 +18,7 @@ namespace Microsoft.OData.ConnectedService.Models
         private ConnectedServiceLogger logger;
 
         [DataMember]
-        public ObservableCollection<string> MruEndpoints { get; private set; }
+        public ObservableCollection<string> MruEndpoints { get; set; }
 
         [DataMember]
         public string ServiceName { get; set; }
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -80,6 +80,9 @@ namespace Microsoft.OData.ConnectedService.Models
         [DataMember]
         public bool IncludeCustomHeaders { get; set; }
 
+        [DataMember]
+        public List<string> ExcludedOperationImports { get; set; }
+
         public UserSettings()
         {
             this.MruEndpoints = new ObservableCollection<string>();
--- a/src/ODataConnectedServiceWizard.cs
+++ b/src/ODataConnectedServiceWizard.cs
@@ -26,101 +26,43 @@ namespace Microsoft.OData.ConnectedService
 
         public ODataConnectedServiceInstance ServiceInstance => this.serviceInstance ?? (this.serviceInstance = new ODataConnectedServiceInstance());
 
-
         public Version EdmxVersion => this.ConfigODataEndpointViewModel.EdmxVersion;
 
         public UserSettings UserSettings { get; }
+
+        private readonly ServiceConfigurationV4 _serviceConfig;
+
+        private string _processedEndpointForOperationImports;
 
         public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
         {
             this.Context = context;
             this.UserSettings = UserSettings.Load(context.Logger);
 
+            // Since ServiceConfigurationV4 is a derived type of ServiceConfiguration. So we can deserialize a ServiceConfiguration into a ServiceConfigurationV4.
+            this._serviceConfig = this.Context.GetExtendedDesignerData<ServiceConfigurationV4>();
+
             ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.UserSettings, this);
             AdvancedSettingsViewModel = new AdvancedSettingsViewModel(this.UserSettings);
-            OperationImportsViewModel = new OperationImportsViewModel();
+            OperationImportsViewModel = new OperationImportsViewModel(this.UserSettings);
 
-            ServiceConfigurationV4 serviceConfig = null;
-
-            OperationImportsViewModel.PageEntering += ObjectSelectionViewModel_PageEntering;
+            OperationImportsViewModel.PageEntering += OperationImportsViewModel_PageEntering;
 
             if (this.Context.IsUpdating)
             {
-                //Since ServiceConfigurationV4 is a derived type of ServiceConfiguration. So we can deserialize a ServiceConfiguration into a ServiceConfigurationV4.
-                serviceConfig = this.Context.GetExtendedDesignerData<ServiceConfigurationV4>();
-                ConfigODataEndpointViewModel.Endpoint = serviceConfig.Endpoint;
-                ConfigODataEndpointViewModel.EdmxVersion = serviceConfig.EdmxVersion;
-                ConfigODataEndpointViewModel.ServiceName = serviceConfig.ServiceName;
-                ConfigODataEndpointViewModel.CustomHttpHeaders = serviceConfig.CustomHttpHeaders;
+                ConfigODataEndpointViewModel.Endpoint = this._serviceConfig.Endpoint;
+                ConfigODataEndpointViewModel.EdmxVersion = this._serviceConfig.EdmxVersion;
+                ConfigODataEndpointViewModel.ServiceName = this._serviceConfig.ServiceName;
+                ConfigODataEndpointViewModel.CustomHttpHeaders = this._serviceConfig.CustomHttpHeaders;
 
-
+                // Restore the main settings to UI elements.
+                ConfigODataEndpointViewModel.PageEntering += ConfigODataEndpointViewModel_PageEntering;
 
                 // The ViewModel should always be filled otherwise if the wizard is completed without visiting this page the generated code becomes wrong
-                AdvancedSettingsViewModel.UseNamespacePrefix = serviceConfig.UseNamespacePrefix;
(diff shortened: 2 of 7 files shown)


unchase:
@habbes @marabooy @KenitoInc @ElizabethOkerio @gathogojr @paulodero it is ready for review

ElizabethOkerio:
@unchase is this an improvement to how PR#50 handles wizard page state or this is completely different?

unchase:
> @unchase is this an improvement to how PR#50 handles wizard page state or this is completely different?

@ElizabethOkerio It's different.

KenitoInc:
@unchase Rebase and solve conflicts so that I can test this PR

unchase:
> @unchase Rebase and solve conflicts so that I can test this PR

@KenitoInc done!

KenitoInc:
@unchase Selections in the Settings page are not persisted. Kindly check.

marabooy:
@KenitoInc  the options seem to work fine but the advanced settings view doesn't get auto expanded. @unchase maybe we could have this if the user expands it then on coming back the pane is still expanded.

unchase:
> @unchase Selections in the Settings page are not persisted. Kindly check.

There are two cases:
1) To generate a new client code
2) To update the generated client code

In the first case:
save_state_new

In the second:
save_state_update

Data is restored when leaving wizard pages.

All works fine.

unchase:
> @KenitoInc the options seem to work fine but the advanced settings view doesn't get auto expanded. @unchase maybe we could have this if the user expands it then on coming back the pane is still expanded.

@marabooy This can be done in a separate PR (not this one).

unchase:
@KenitoInc @habbes @paulodero @gathogojr can you review it?

unchase:
> @unchase This PR seems to no longer validate and parse the service endpoint when leaving the service endpoint page. This allows the user to proceed to subsequent pages without providing a service endpoint. The expected behavior is to prevent going passed the endpoint page without a valid edm model, an error should instead be shown if the service endpoint is empty, can't be fetched or can't be parsed.

@habbes This bug appeared along with PR #50 (not related to my PR), but I will try to fix it.

unchase:
@habbes Done! Now it works fine. Also fixed the issue #94
Please, check it.

habbes:
@unchase now when I create a new service, the UI is preloaded with the settings of the last created service. The wizard should be initialized with default settings when creating a new service.

unchase:
> @unchase now when I create a new service, the UI is preloaded with the settings of the last created service. The wizard should be initialized with default settings when creating a new service.

@habbes Aren't `UserSettings` meant for this? Or do it has some other purpose?

habbes:
@unchase we decide that for now creating a new service will always use the default settings. As a result we decided not to load the prelaoad the UI from user settings. If the user wants to re-use existing config, they would need to upload the json config file.

habbes:
If based on feedback, it's clear that users prefer for the settings to be persisted and be automatically applied to the next connected service, it will be relatively easy to implement that change. However, going from persisting to not persisting would be a more destructive change. That's why we decided to go with this approach first.

unchase:
> If based on feedback, it's clear that users prefer for the settings to be persisted and be automatically applied to the next connected service, it will be relatively easy to implement that change. However, going from persisting to not persisting would be a more destructive change. That's why we decided to go with this approach first.

@habbes Ok, I commented out all the lines `LoadFromUserSettings();`. If will need to change the approach, just will should to uncomment them.

habbes:
@unchase  in my opinion it's better to remove code than to comment it out. Especially when it's not hard to rewrite the code if it's needed. The code around it might change a lot such that uncommenting it alone might not be sufficient for it to work. And people may be afraid to touch/remove commented code even if it's not needed anymore.

unchase:
> @unchase in my opinion it's better to remove code than to comment it out. Especially when it's not hard to rewrite the code if it's needed. The code around it might change a lot such that uncommenting it alone might not be sufficient for it to work. And people may be afraid to touch/remove commented code even if it's not needed anymore.

@habbes i am agree. Done.

unchase:
> @unchase I've tested again. It works fine when you fill in the fields manually. However, it has issues when you load values from a config file:
> 
> * it seems that the advanced settings are not loaded from the config file. When you navigate to that page, the fields are empty or default even if you had set those values in the config file
> * When you load values from a config file then navigate to the second page and back to the first page, the values in the service name and web proxy will have reset to their default or empty, like it does not properly save page state in these cases

@habbes fixed.

ElizabethOkerio:
This PR breaks some of the functionalities of the load configs feature that was merged. For instance, if I load configs from a json file and the service name that is loaded is for instance "MyService". I expect the generated service folder to be "MyService" but in this case it loads "Odata Service". The default name. Most of the first page configs don't get loaded as expected if loaded from a json file. Please look at this.

habbes:
@ElizabethOkerio it seemed I had shared the wrong build for testing, hence the issues you encountered.

@unchase  I've tested again and noticed that it tries to fetch the service model the moment that you load the config file, instead of waiting until I click "Next" to leave the config endpoint page. This means that if the config file does not have a service endpoint, then it will not be loaded because the connected service will fail to fetch the metadata. Instead you will get an error message.

unchase:
> @ElizabethOkerio it seemed I had shared the wrong build for testing, hence the issues you encountered.
> 
> @unchase I've tested again and noticed that it tries to fetch the service model the moment that you load the config file, instead of waiting until I click "Next" to leave the config endpoint page. This means that if the config file does not have a service endpoint, then it will not be loaded because the connected service will fail to fetch the metadata. Instead you will get an error message.

@habbes I fixed it.

weitzhandler:
@unchase 

> Fix issues #69 #94

Consider using [GitHub keywords](https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords), may instead write `fix #69, fix #94`.

KenitoInc:
> @ElizabethOkerio it seemed I had shared the wrong build for testing, hence the issues you encountered.
> @unchase I've tested again and noticed that it tries to fetch the service model the moment that you load the config file, instead of waiting until I click "Next" to leave the config endpoint page. This means that if the config file does not have a service endpoint, then it will not be loaded because the connected service will fail to fetch the metadata. Instead you will get an error message.
> 
> @habbes I fixed it.

@unchase The config is loaded, though the error popup message is still there

unchase:
> > @ElizabethOkerio it seemed I had shared the wrong build for testing, hence the issues you encountered.
> > @unchase I've tested again and noticed that it tries to fetch the service model the moment that you load the config file, instead of waiting until I click "Next" to leave the config endpoint page. This means that if the config file does not have a service endpoint, then it will not be loaded because the connected service will fail to fetch the metadata. Instead you will get an error message.
> > @habbes I fixed it.
> 
> @unchase The config is loaded, though the error popup message is still there

@KenitoInc That is, do not need to display an error?

KenitoInc:
popup
@unchase I am getting this pop up when I don't have an endpoint in the configuration file. This should not appear until I click Next

unchase:
> popup
> 
> @unchase I am getting this pop up when I don't have an endpoint in the configuration file. This should not appear until I click Next

@KenitoInc please, check it now.

KenitoInc:
@unchase One more request. Am asking for permission to use the gif (in the PR description) in the blog post for the next release.

unchase:
> @unchase One more request. Am asking for permission to use the gif (in the PR description) in the blog post for the next release.

@KenitoInc Yes of course, use it.

Issue #54, opened 2020-03-22
File Access Error occurs when generating client using OData Connected Service 0.6.0


habbes:
### Describe the bug
A file access error occurs when generating code, and the created Csdl.xml file is empty.

#### Version of the Project affected
ODataConnectedService 0.6.0

#### To Reproduce
1. Add a new OData Connected Service to your project
2. Provide a service endpoint url
3. Click Finish

#### Expected behavior
It should generated client code and Csdl.xml file containing the data model

#### Actual behavior
- It fails with an error saying it cannot access the Csdl.xml file.
- It creates a Csdl.xml file but it's empty

#### Screenshots

![MicrosoftTeams-image](https://user-images.githubusercontent.com/8460169/77259681-55489080-6c94-11ea-9257-d0dad9cff699.png)

#### Additional context

Bug was most likely introduced by PR https://github.com/OData/ODataConnectedService/pull/46

It appears that the path of the Csdl.xml that causes the error varies from machine to machine. In the screenshot above it seems to be in system32. I tested on a different machine and the path was that of the Visual Studio installation.

I've added my investigation and proposed solution in the comments below.

habbes:
This error does not seem to occur when you're debugging in VS experimental version. However, even then, it appears that the Csdl.xml file that's generated is actually empty. And when I try to run a console app that uses the generated client code, I get a `FileNotFoundException`.

habbes:
I've done a bit of investigation and it appears that some, if not all, of the highlighted problems are due to the use of a relative file path for the `Csdl.xml` file. Since only the file name is given for that file, its path is resolved based on the "current" directory. However, the "current directory" is really context-specific. The current directory when the connected service is running during code generation may be different from when the project itself is running, it may be different depending on whether you're debugging the extension, or whether you've downloaded and installed the vsix, it may be different from machine to machine or user to user depending on your environment and configuration.

In one of the test runs I've just done, when the code was being generated, the `Csdl.xml` file is created using an absolute path based on the connected service reference folder (i.e. `tempMetadataFile`), this is what is visible in solution explorer under the connected service folder and it appears to be `projectRoot\Connected Services\Service Name\Csdl.xml`. However, this is not the path that's provided to the `t4CodeGenerator.TempFilePath` property, this property is just given the simple relative file name `Csdl.xml`.

[...]

PR #86, opened 2020-04-01
Make temp file metadata an embedded resource

KenitoInc:
Fixes Issue: #54 
_Background_
The current directory when the connected service is running during code generation may be different from when the project itself is running. Therefore we need a way for the generated Csdl.xml to be accessed by the connected service.
Using `relative path` requires us to handle the paths to the working directory for project and connected service which create additional complexity.
`Absolute path` would solve this. The main drawback is that when you move the project folder to another folder, we will need to regenerate the files afresh since the path will be invalid.

**Solution**
We make the Csdl.xml an embedded resource which will be shipped with the executable. With this we will not have to worry about paths.

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -12,6 +12,7 @@ using EnvDTE;
 using Microsoft.OData.ConnectedService.Models;
 using Microsoft.OData.ConnectedService.Templates;
 using Microsoft.VisualStudio.ConnectedServices;
+using VSLangProj;
 
 namespace Microsoft.OData.ConnectedService.CodeGeneration
 {
@@ -83,6 +84,11 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
 
             tempFile = Path.GetTempFileName();
 
+            var metadataFile = Path.Combine(referenceFolder, Common.Constants.CsdlFileName);
+            await this.Context.HandlerHelper.AddFileAsync(tempFile, metadataFile, new AddFileOptions() { SuppressOverwritePrompt = true });
+            var projectItem = this.GetCsdlFileProjectItem(Common.Constants.CsdlFileName);
+            projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
+
             using (StreamWriter writer = File.CreateText(tempFile))
             {
                 var text = File.ReadAllText(Path.Combine(t4Folder, "ODataT4CodeGenerator.tt"));
@@ -106,6 +112,8 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
                 text = Regex.Replace(text, "(public const bool MakeTypesInternal = )false;", "$1" + ServiceConfiguration.MakeTypesInternal.ToString().ToLower(CultureInfo.InvariantCulture) + ";");
                 var customHeaders = ServiceConfiguration.CustomHttpHeaders ?? "";
                 text = Regex.Replace(text, "(public const string CustomHttpHeaders = )\"\";", "$1@\"" + customHeaders + "\";");
+                text = Regex.Replace(text, "(public const string MetadataFilePath = )\"\";", "$1@\"" + metadataFile + "\";");
+                text = Regex.Replace(text, "(public const string MetadataFileRelativePath = )\"\";", "$1@\"" + Common.Constants.CsdlFileName + "\";");
                 await writer.WriteAsync(text);
                 await writer.FlushAsync();
             }
@@ -153,6 +161,23 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
             t4CodeGenerator.WebProxyNetworkCredentialsDomain = ServiceConfiguration.WebProxyNetworkCredentialsDomain;
 
             var tempFile = Path.GetTempFileName();
+            var referenceFolder = GetReferenceFileFolder();
+
+            var metadataFile = Path.Combine(referenceFolder, Common.Constants.CsdlFileName);
+            await this.Context.HandlerHelper.AddFileAsync(tempFile, metadataFile, new AddFileOptions() { SuppressOverwritePrompt = true });
+
+            // Hack!
+            // Tests were failing since the test project cannot access ProjectItems
+            // dte == null when running test cases
+            var dte = VisualStudio.Shell.Package.GetGlobalService(typeof(EnvDTE.DTE)) as EnvDTE.DTE;
+            if(dte != null)
+            {
+                var projectItem = this.GetCsdlFileProjectItem(Common.Constants.CsdlFileName);
+                projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
+            }
+
+            t4CodeGenerator.MetadataFilePath = metadataFile;
+            t4CodeGenerator.MetadataFileRelativePath = Common.Constants.CsdlFileName;
 
             using (StreamWriter writer = File.CreateText(tempFile))
             {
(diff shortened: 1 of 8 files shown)


src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -153,6 +162,24 @@ private async Task AddGeneratedCodeAsync()
             t4CodeGenerator.WebProxyNetworkCredentialsDomain = ServiceConfiguration.WebProxyNetworkCredentialsDomain;
 
             string tempFile = Path.GetTempFileName();
+            const string CsdlFileName = "Csdl.xml";
+            string referenceFolder = GetReferenceFileFolder();
+
+            var tempMetadataFile = Path.Combine(referenceFolder, CsdlFileName);
+            await this.Context.HandlerHelper.AddFileAsync(tempFile, tempMetadataFile);
```
habbes, line 169: I think you should add the `SuppressOverwritePrompt` so that it doesn't ask the user whether to overwrite the Csdl.xml file when updating the connected service.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -415,6 +420,15 @@ public string TempFilePath
     set;
 }
 
+/// <summary>
+/// The relative path for the temporary file where the metadata xml document can be stored.
+/// </summary>
+public string TempFileRelativePath
```
gathogojr, line 426: @KenitoInc This can a one liner. I don't think there's any value in having the property span 5 lines

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1151,6 +1166,15 @@ public string TempFilePath
         set;
     }
 
+    /// <summary>
+    /// The relative path for the temporary file where the metadata xml document can be stored.
+    /// </summary>
+    public string TempFileRelativePath
```
gathogojr, line 1172: https://github.com/OData/ODataConnectedService/pull/86/files#r410016895

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -3836,6 +3857,7 @@ namespace <#= fullNamespace #>
     internal override void WriteGeneratedEdmModel(string escapedEdmxString)
     {
         string path = this.context.TempFilePath;
+        string relativePath = this.context.TempFileRelativePath;
```
gathogojr, line 3860: var?

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -275,6 +277,15 @@ public string TempFilePath
     set;
 }
 
+/// <summary>
+/// The relative path for the temporary file where the metadata xml document can be stored.
```
gathogojr, line 281: @KenitoInc Like @habbes observed where it talks about temp/temporary file I think you should adjust that to "metadata file"

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -3982,7 +4004,10 @@ namespace <#= fullNamespace #>
                 [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "<#=T4Version#>")]
                 private static global::System.Xml.XmlReader CreateXmlReader()
                 {
-                    return global::System.Xml.XmlReader.Create(new global::System.IO.StreamReader(filePath));
+                    var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
+                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
```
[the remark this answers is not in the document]
KenitoInc: In this case both work as expected. First() will throw if it cannot find the first matching value, Single() will throw if it cannot find the value and if there are more than one matching element in the input sequence. We cannot have more than one element with the same file path.

src/CodeGeneration/V4CodeGenDescriptor.cs
KenitoInc: From the docs, it's not supported in VS 2019. I will add it though https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.connectedservices.addfileoptions.suppressoverwriteprompt?view=visualstudiosdk-2017

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I kept in consistent with the other properties in the tt file

src/Templates/ODataT4CodeGenerator.ttinclude
KenitoInc: I did not want to mix types and var within the tt files, we can rename all of them later

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -153,6 +162,24 @@ private async Task AddGeneratedCodeAsync()
             t4CodeGenerator.WebProxyNetworkCredentialsDomain = ServiceConfiguration.WebProxyNetworkCredentialsDomain;
 
             string tempFile = Path.GetTempFileName();
+            const string CsdlFileName = "Csdl.xml";
```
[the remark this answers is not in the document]

src/Templates/ODataT4CodeGenerator.ttinclude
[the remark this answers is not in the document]
KenitoInc: #Resolved

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -67,6 +68,7 @@ private async Task AddT4FileAsync()
             var t4Folder = Path.Combine(this.CurrentAssemblyPath, "Templates");
 
             var referenceFolder = GetReferenceFileFolder();
+            const string csdlFileName = "Csdl.xml";
```
paulodero, line 71: For constants, PascalCasing is recommended.

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -152,6 +161,24 @@ private async Task AddGeneratedCodeAsync()
             t4CodeGenerator.WebProxyNetworkCredentialsDomain = ServiceConfiguration.WebProxyNetworkCredentialsDomain;
 
             var tempFile = Path.GetTempFileName();
+            const string csdlFileName = "Csdl.xml";
```
paulodero, line 164: You could move constants in the `Constants.cs` file to avoid duplication

src/CodeGeneration/V4CodeGenDescriptor.cs
KenitoInc: #Resolved

KenitoInc: #Resolved

src/CodeGeneration/V4CodeGenDescriptor.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -3982,7 +4004,10 @@ namespace <#= fullNamespace #>
                 [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "<#=T4Version#>")]
                 private static global::System.Xml.XmlReader CreateXmlReader()
```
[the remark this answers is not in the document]
KenitoInc: Nope, when you look at the entire code, you will see that it's properly indented

src/Templates/ODataT4CodeGenerator.ttinclude
KenitoInc: #Resolved

src/CodeGeneration/V4CodeGenDescriptor.cs

src/Templates/ODataT4CodeGenerator.ttinclude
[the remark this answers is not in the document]
KenitoInc: #Resolved

PR #87, opened 2020-04-03
Auto selects last metadata endpoint

weitzhandler:
Hi,

The endpoint drop-down contains the recent endpoints that were used.

image

When working on a server-client app that requires frequent changes to the server and re-generation of the client entities, any redundant click is tedious.
This PR changes the `ComboBox` to auto-select the first URL.

### DIFF
--- a/src/Views/ConfigODataEndpoint.xaml
+++ b/src/Views/ConfigODataEndpoint.xaml
@@ -12,6 +12,7 @@
     d:DataContext="{d:DesignInstance Type=viewModels:ConfigODataEndpointViewModel}"
     d:DesignHeight="300"
     d:DesignWidth="500"
+	Loaded="UserControl_Loaded"
     mc:Ignorable="d">
     <UserControl.Resources>
         <BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter" />
--- a/src/Views/ConfigODataEndpoint.xaml.cs
+++ b/src/Views/ConfigODataEndpoint.xaml.cs
@@ -27,7 +27,7 @@ namespace Microsoft.OData.ConnectedService.Views
 
         private void UserControl_Loaded(object sender, RoutedEventArgs e)
         {
-            if (Endpoint.Items.Count > 0)
+            if (Endpoint.Items.Count > 0 && !((ConfigODataEndpointViewModel)this.DataContext).ServiceWizard.Context.IsUpdating)
             {
                 Endpoint.SelectedItem = Endpoint.Items[0];
             }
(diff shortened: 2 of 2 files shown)


KenitoInc:
@weitzhandler Rebase your branch and fix conflicts. Once your PR is ready for review, inform us.

weitzhandler:
@KenitoInc conflicts resolved.

KenitoInc:
> @KenitoInc conflicts resolved.

Is the PR ready for review?

weitzhandler:
As far as it's up to me, yes.

KenitoInc:
@weitzhandler Rebase so that we can prioritize this PR for the next release

weitzhandler:
Can you please explain? My branch is based on upstream/master

gathogojr:
@weitzhandler Kindly rebase

src/Views/ConfigODataEndpoint.xaml
```diff
@@ -12,6 +12,7 @@
     d:DataContext="{d:DesignInstance Type=viewModels:ConfigODataEndpointViewModel}"
     d:DesignHeight="300"
     d:DesignWidth="500"
+	Loaded="UserControl_Loaded"
```
gathogojr, line 15: @weitzhandler Just wondering, would this "Loaded" be fired in an update scenario? 'cause we wouldn't want it selecting an endpoint arbitrarily in that scenario

weitzhandler: The endpoint dropdown list is supposed to be ordered, so this selects the first available. Can that be a problem?

gathogojr: @weitzhandler If you're update the connected service, the endpoint is loaded from the settings file. I just wanted to know that the endpoint loaded from the settings file is not replaced by the last used endpoint in such a scenario

weitzhandler: > the endpoint is loaded from the settings file

Where is it loaded to? Is the value added to the `ComboBox` or just as text?

weitzhandler:
Do you think it'd be better to have a method "Select first entry", which we call on loading, and after loading settings, and maybe after some other important actions?
We could also hook it up to the `ItemsSource` changed of the combo.

gathogojr:
@weitzhandler How about something like this in your `UserControl_Loaded` method:
```
if (!((ConfigODataEndpointViewModel)this.DataContext).ServiceWizard.Context.IsUpdating && Endpoint.Items.Count > 0)
{
    Endpoint.SelectedItem = Endpoint.Items[0];
}
```

weitzhandler:
Thank you John for your commit.

gathogojr:
> Thank you John for your commit.
> Does it still need work?

Verified and it works now. It's ready to be merged

PR #109, opened 2020-04-21
fix Schema types and operation imports naming

### LINKED ISSUE #108: Harmonize ordering and naming of Schema Types and Function/Action Imports wizard pages
### Issue description
1. Currently, if say I have a function import _GetNearestAirport_ as the only object that depends on _Airport_ entity type, and I don't need either to be emitted on the Connected Service proxy, I'm forced to first skip **Schema Types** wizard page, go to the **Function/Action Imports** wizard page and uncheck it, before going back to the **Schema Types** wizard page to uncheck _Airport_
2. The title on the **Schema Types** wizard page is **Schema Selection** while the title on the **Function/Action Imports** wizard page is **Object Selection**

#### Proposed solution
1. Since **Function/Action Imports** depends on **Schema Types** and not the other way around, I believe the logical thing would be to have **Function/Action Imports** wizard page appear above **Schema Types** wizard page. That way, if I uncheck a function or action import on the **Function/Action Imports** wizard page, then as I proceed to the **Schema Types** wizard page I'll not having any trouble unchecking the schema types it depends on in a scenario similar to the one described above
2. Harmonize the naming:-
   - Title for **Schema Types** wizard page could be **Schema Types Selection**
   - Title for **Function/Action Imports** wizard page could be **Function/Action Imports Selection**
   - Or any suitable naming so long as its harmonious

N/A

#### Additional context
N/A

paulodero:
Fixes #108

### DIFF
--- a/src/ViewModels/OperationImportsViewModel.cs
+++ b/src/ViewModels/OperationImportsViewModel.cs
@@ -20,9 +20,9 @@ namespace Microsoft.OData.ConnectedService.ViewModels
 
         public OperationImportsViewModel(UserSettings userSettings = null) : base()
         {
-            Title = "Object Selection";
+            Title = "Function/Action Imports";
             Description = "Select function and action imports to include in the generated code.";
-            Legend = "Function/Action Imports";
+            Legend = "Function/Action Imports Selection";
             OperationImports = new List<OperationImportModel>();
             IsSupportedODataVersion = true;
             this.UserSettings = userSettings;
--- a/src/ViewModels/SchemaTypesViewModel.cs
+++ b/src/ViewModels/SchemaTypesViewModel.cs
@@ -38,9 +38,9 @@ namespace Microsoft.OData.ConnectedService.ViewModels
 
         public SchemaTypesViewModel(UserSettings userSettings = null) : base()
         {
-            Title = "Schema selection";
+            Title = "Schema Types";
             Description = "Select schema types to include in the generated code.";
-            Legend = "Schema Types";
+            Legend = "Schema Types Selection";
             SchemaTypes = new List<SchemaTypeModel>();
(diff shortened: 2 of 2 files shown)


src/packages.config
```diff
@@ -1,10 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="EnvDTE" version="8.0.2" targetFramework="net472" />
-  <package id="Microsoft.OData.Client" version="7.6.3" targetFramework="net472" />
```
marabooy, line 4: Not sure these are needed

test/ODataConnectedService.Tests/ODataConnectedService.Tests.csproj
```diff
@@ -76,7 +76,7 @@
       <Version>2.0.0.1</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.OData.Client">
-      <Version>7.6.3</Version>
```
marabooy, line 79: Same here

src/packages.config
paulodero: Is it a problem to reference the latest OData Client?

marabooy: I would probably keep it at a minimum version. been seeing people having some regressions with this release. two new bugs got added on odata.net so we may need to keep the old version. I know currently we havent upgraded the current referenced libs in the build. Maybe we should have one version both in t4 and in this.

paulodero: Reverted.

test/ODataConnectedService.Tests/ODataConnectedService.Tests.csproj
paulodero: Reverted.

PR #110, opened 2020-04-23
Adding test cases for ConfigODataEndPointViewModel Class

paulodero:
### ConfigODataEndpointViewModel

Handles the endpoint config page of the wizard, also fetches the model from the provided endpoint.
`OnPageLeavingAsync()` is called when the page is leaving, and it handles a lot of logic that's not tested:

- [ ] Should add the provided endpoint to the user setting's most-recently used list when leaving the page
- [x] Should fetch metadata when from the provided endpoint when leaving the page and save it to a temporary file
- [x] Should detect and set the edmx version from the metadata
- [x] Should be able to fetch metadata if the endpoint is a local file
- [ ] Should pass provided custom headers to the web request when fetching metadata from a service
- [ ] Should be able to fetch metadata when behind web proxy

### DIFF
--- a/src/ODataConnectedServiceWizard.cs
+++ b/src/ODataConnectedServiceWizard.cs
@@ -44,10 +44,10 @@ namespace Microsoft.OData.ConnectedService
         public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
         {
             this.Context = context;
-            this.UserSettings = UserSettings.Load(context.Logger);
+            this.UserSettings = UserSettings.Load(context?.Logger);
 
             // Since ServiceConfigurationV4 is a derived type of ServiceConfiguration. So we can deserialize a ServiceConfiguration into a ServiceConfigurationV4.
-            this._serviceConfig = this.Context.GetExtendedDesignerData<ServiceConfigurationV4>();
+            this._serviceConfig = this.Context?.GetExtendedDesignerData<ServiceConfigurationV4>();
 
             ConfigODataEndpointViewModel = new ConfigODataEndpointViewModel(this.UserSettings, this);
             AdvancedSettingsViewModel = new AdvancedSettingsViewModel(this.UserSettings);
@@ -57,7 +57,7 @@ namespace Microsoft.OData.ConnectedService
             OperationImportsViewModel.PageEntering += OperationImportsViewModel_PageEntering;
 
             SchemaTypesViewModel.PageEntering += SchemaTypeSelectionViewModel_PageEntering;
-            if (this.Context.IsUpdating)
+            if (this.Context != null && this.Context.IsUpdating)
             {
                 ConfigODataEndpointViewModel.Endpoint = this._serviceConfig.Endpoint;
                 ConfigODataEndpointViewModel.EdmxVersion = this._serviceConfig.EdmxVersion;
--- a/test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
+++ b/test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
@@ -1,0 +1,82 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.OData.ConnectedService;
+using Microsoft.OData.ConnectedService.Common;
+using Microsoft.OData.ConnectedService.Models;
+using Microsoft.OData.ConnectedService.Tests.Templates;
+using Microsoft.OData.ConnectedService.ViewModels;
+using Microsoft.VisualStudio.ConnectedServices;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace ODataConnectedService.Tests.ViewModels
+{
+    [TestClass]
+    public class ConfigOdataEndPointViewModelTests
+    {
+        private static ConfigODataEndpointViewModel configOdataEndPointViewModel;
+        private static UserSettings userSettings;
+        private static ODataConnectedServiceWizard serviceWizard;
+
+        [TestInitialize]
+        public void Init()
+        {
+            userSettings = new UserSettings();
+            serviceWizard = new ODataConnectedServiceWizard(null);
+            configOdataEndPointViewModel = new ConfigODataEndpointViewModel(userSettings, serviceWizard);
+        }
+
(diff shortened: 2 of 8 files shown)


test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            pageNavigationResult = pageNavigationResultTask?.Result;
+            Assert.IsNotNull(pageNavigationResult.ErrorMessage);
+            Assert.IsTrue(pageNavigationResult.ErrorMessage.Contains(Constants.InputServiceEndpointMsg),"User is not prompted to enter endpoint");
+            Assert.IsFalse(pageNavigationResult.IsSuccess);
+            Assert.IsTrue(pageNavigationResult.ShowMessageBoxOnFailure);
+
+            //Provide a url without $metadata
+            configOdataEndPointViewModel.Endpoint = "http://localhost/ODataServivce";
```
KenitoInc, line 51: `http://localhost/ODataService` would have been better

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            //Check if $metadata is apended if the url does not have it added at the end
+            Assert.AreEqual(configOdataEndPointViewModel.Endpoint, "http://localhost/ODataServivce/$metadata");
```
KenitoInc, line 55: `http://localhost/ODataService/$metadata` would have been better

src/source.extension.vsixmanifest
```diff
@@ -1,27 +1,27 @@
...
-        <Dependency Id="Microsoft.Framework.NDP" DisplayName="Microsoft .NET Framework" d:Source="Manual" Version="[4.5,)" />
-    </Dependencies>
-    <Assets>
-        <Asset Type="Microsoft.VisualStudio.MefComponent" d:Source="Project" Path="|%CurrentProject%|" />
-    </Assets>
-    <Prerequisites>
-        <Prerequisite Id="Microsoft.VisualStudio.Component.CoreEditor" Version="[15.0,17.0)" DisplayName="Visual Studio core editor" />
-    </Prerequisites>
+  <Metadata>
+    <Identity Id="ODataConnectedService.e30335d6-f9c7-4d08-b66a-f011f3f18477" Version="0.9.1" Language="en-US" Publisher="Microsoft" />
```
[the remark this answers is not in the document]
paulodero: version already increased to v0.9.1. Will rebase to remove this.

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
paulodero: I don't understand. Please clarify

KenitoInc: `ODataServivce` looks like a typo

paulodero: Got it. Thanks

src/ODataConnectedServiceWizard.cs
```diff
@@ -42,10 +42,10 @@ internal class ODataConnectedServiceWizard : ConnectedServiceWizard
         public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
         {
             this.Context = context;
-            this.UserSettings = UserSettings.Load(context.Logger);
+            this.UserSettings = UserSettings.Load(context?.Logger);
```
marabooy, line 45: Shouldn't we just make sure this always is never null. if null initialize a different logger.

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        [TestInitialize]
+        public void Init()
+        {
+            userSettings = new UserSettings();
+            serviceWizard = new ODataConnectedServiceWizard(null);
+            configOdataEndPointViewModel = new ConfigODataEndpointViewModel(userSettings, serviceWizard);
+        }
+
+        [TestMethod]
+        public void OnPageLeavingConfigODataEndpointPageTest()
```
marabooy, line 32: Probably also good that throws since it is configured with gibberish.

src/ODataConnectedServiceWizard.cs
habbes: In my opinion it's okay if we at least allow the logger to be nullable so that we don't have to mock `logger.WriteMessageAsync` since it doesn't add as much value to the testing. But I think we do already have tests that have mocked it. I also have a dummy subclass for `ConnectedServiceProviderContext` in my (unmerged) tests, which may or may not be useful here to avoid nulling the context.

ElizabethOkerio:
@paulodero rebase and fix conflicts

src/ODataConnectedServiceWizard.cs
paulodero: @marabooy For example which logger in this case?

paulodero:
> @paulodero rebase and fix conflicts

@ElizabethOkerio Done.

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
paulodero: Fixed.

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
paulodero: Fixed.

src/source.extension.vsixmanifest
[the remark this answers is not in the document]
paulodero: Rebased.

src/ODataConnectedServiceWizard.cs
marabooy: A mocked version. My preference is to avoid nulls.

PR #116, opened 2020-04-29
Fix #115: MetadataDocumentUri not properly escaping local file paths when generating T4 templates

### LINKED ISSUE #115: MetadataDocumentUri not properly escaping local file paths when generating T4 templates
### Describe the bug
When you generated a connected service with a local file path for the metadata document and then opt to generate the T4 templates instead of the actual code, then the `\` separator in the local path in generated `Reference.tt` file is not escaped. This causes errors when you run the T4 custom tool.

ODataConnectedService v0.9.0

#### To Reproduce
1. Create an OData Connected Service
2. Use a local file as the service document
3. In the advanced settings, opt to generate T4 templates
4. Finish the code generation

#### Expected behavior
The `MetadataDocumentUri` setting in the generated `Reference.tt` file should be properly escaped or use verbatim strings, e.g.:
```c#
public const string MetadataDocumentUri = @"C:\Users\TestVM\Downloads\Temp\Trippin.xml";
```

#### Actual behavior
The path is generated without the backslash separators being escaped:
```c#
public const string MetadataDocumentUri = "C:\Users\TestVM\Downloads\Temp\Trippin.xml";
```

#### Additional context
The user can manually edit the path in directly in the `Reference.tt` configuration file, and just prefix the string with `@`, so this is definitely not a blocker. However, since the template does not produced useful error messages when it fails, it's good to ensure the generate Reference.tt works out of the box. Furthermore, the `MetadataFilePath` setting has the `@` on the string literal, so shouldn't be a sweat to fix.

unchase:
Fix #115

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -99,7 +99,7 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
                 var text = File.ReadAllText(Path.Combine(t4Folder, "ODataT4CodeGenerator.tt"));
 
                 text = Regex.Replace(text, "ODataT4CodeGenerator(\\.ttinclude)", this.GeneratedFileNamePrefix + "$1");
-                text = Regex.Replace(text, "(public const string MetadataDocumentUri = )\"\";", "$1\"" + ServiceConfiguration.Endpoint + "\";");
+                text = Regex.Replace(text, "(public const string MetadataDocumentUri = )\"\";", "$1@\"" + ServiceConfiguration.Endpoint + "\";");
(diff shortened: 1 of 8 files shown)


unchase:
> Could you add tests for this?

Done (fixed some files).
Test case is handled by `TestAddGeneratedClientCode_GeneratesT4TemplateFiles` test.

unchase:
> @unchase Run custom tool on `ODataT4CodeGenerator.tt` file and commit the generated `ODataT4CodeGenerator.cs` file

@KenitoInc done!

habbes:
@unchase could you rebase this branch then we merge it.

Also, it appears that you use an auto-formatter for your code. Is this the case? If so, which auto-formatting tool do you use?

unchase:
> @unchase could you rebase this branch then we merge it.
> 
> Also, it appears that you use an auto-formatter for your code. Is this the case? If so, which auto-formatting tool do you use?

@habbes i use [Trailing Whitespace Visualizer](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer) by Mads Kristensen.

Rebased.

PR #117, opened 2020-04-29
Fix #114: Unselecting all schema types does not deselect dependent function imports

### LINKED ISSUE #114: Unselecting all schema types does not deselect dependent function imports
### Describe the bug
When you deselect all schema types, function/action imports that depend on the deselected types can still remain selected, and if you proceed to generate code, it will not compile.

ODataConnectedService v0.9.0

#### To Reproduce
1. Create a new OData connected service using the Trippin sample service (https://services.odata.org/V4/(S(oh2omrh0ktdqbvdgamvhhbec))/TripPinServiceRW/$metadata)
2. Move to the schema types page, all types should be selected by default. 
3. Proceed to the function/action imports page, all items should also be selected by default.
4. Go back to the schema type pages and click "Unselect All" to unselect all types
5. Proceed to the function/action imports page. Notice that all functions are still selected, but all the types are still unselected.
6. If you click Finish at this point, the generated code will have compile-time errors because `GetNearestAirport` function was generated, but the required types like `Airport` were not
7. Repeat steps 1-5. Then unselect the `GetNearestAirport` function. Strangely enough, doing so selects all the types required by the function. You can confirm this by going back to schema types page and find that `Airport` and other types are selected.

#### Expected behavior
- Unselecting all schema types should either:
  - Unselect all functions that require the unselected types
  - or it should show a prompt when you leave the page showing that some unselected types are still dependent on and then auto-select those types needed by the functions.
- Unselecting a function/import should not cause types required by that function to be selected

#### Actual behavior
- Unselecting all schema types does not affect selected function/action imports which depend on the unselected types
- Unselecting a function/import causes schema types needed by that function to be automatically selected

#### Additional context
Unselecting individual schema types from the list seems to work as expected, i.e. when you unselect individual schema types and navigate away from the page, you get a prompt telling you unselected types are required, and then they're auto-selected.

unchase:
Fix #114 :
* when an schema type is deselected, all the types of its Navigational properties are automatically deselected;
* deselect all operations that require the deselected types.

Added the tests below:

To **ODataConnectedServiceWizardTests**:
- [x] `ShouldDeselectOperations_WhenRelatedTypeIsDeselectedBefore`.

To **SchemaTypesViewModelTests**:
- [x] `LoadSchemaTypes_ShouldSetAllSchemaTypesAsSelectedAndOrderedByName`;
- [x] `LoadSchemaTypes_ShouldAddRelatedTypesForStructuredTypesWithBaseType`;
- [x] `SelectSchemaType_ShouldSelectItsRelatedTypes`;
- [x] `DeselectSchemaType_ShouldDeselectRelatedTypes`;
- [x] `ExcludeSchemaTypes_ShouldDeselectTheSpecifiedTypes`;
- [x] `ExcludedSchemaTypeNames_ShouldReturnNamesOfDeselectedTypes`;
- [x] `ClearSchemaTypes_ShouldResetSchemaTypesToAnEmptyCollection`;
- [x] `SelectAllSchemaTypes_ShouldMarkAllTypesAsSelected`;
- [x] `DeselectAllSchemaTypes_ShouldMarkAllTypesAsNotSelected`.

### DIFF
--- a/src/ViewModels/SchemaTypesViewModel.cs
+++ b/src/ViewModels/SchemaTypesViewModel.cs
@@ -36,7 +36,9 @@ namespace Microsoft.OData.ConnectedService.ViewModels
 
         internal bool IsEntered;
 
-        public SchemaTypesViewModel(UserSettings userSettings = null) : base()
+        private readonly ODataConnectedServiceWizard _wizard;
+
+        public SchemaTypesViewModel(UserSettings userSettings = null, ODataConnectedServiceWizard wizard = null) : base()
         {
             Title = "Schema Types";
             Description = "Select schema types to include in the generated code.";
@@ -45,6 +47,7 @@ namespace Microsoft.OData.ConnectedService.ViewModels
             SchemaTypeModelMap = new Dictionary<string, SchemaTypeModel>();
             RelatedTypes = new Dictionary<string, ICollection<string>>();
             this.UserSettings = userSettings;
+            this._wizard = wizard;
         }
 
         /// <summary>
@@ -74,13 +77,29 @@ namespace Microsoft.OData.ConnectedService.ViewModels
 
             var correctTypeSelection = true;
 
+            // exclude related operationimports for excluded types
+            if (this._wizard != null)
+            {
+                IEdmModel model = EdmHelper.GetEdmModelFromFile(this._wizard.ConfigODataEndpointViewModel.MetadataTempPath);
+                IEnumerable<IEdmOperationImport> operations = EdmHelper.GetOperationImports(model);
+                IEnumerable<IEdmOperationImport> operationsToExclude = operations.Where(x => !this._wizard.OperationImportsViewModel.IsOperationImportIncluded(x,
+                    ExcludedSchemaTypeNames.ToList())).ToList();
+                foreach (var operationImport in this._wizard.OperationImportsViewModel.OperationImports)
+                {
+                    if (operationsToExclude.Any(x => x.Name == operationImport.Name))
+                    {
+                        operationImport.IsSelected = false;
+                    }
+                }
+            }
+
             //check each excluded schema type and check if they are required. If so, then automatically select them.
             foreach (var schemaType in ExcludedSchemaTypeNames)
             {
-                if (RelatedTypes.TryGetValue(schemaType, out ICollection<string> relatedtypes))
+                if (RelatedTypes.TryGetValue(schemaType, out ICollection<string> relatedTypes))
                 {
                     //Check if any of the related types has been selected
-                    if (relatedtypes.Any(o =>
+                    if (relatedTypes.Any(o =>
                     {
                         if(SchemaTypeModelMap.TryGetValue(o, out SchemaTypeModel schemaTypeModel))
                         {
@@ -188,7 +207,7 @@ namespace Microsoft.OData.ConnectedService.ViewModels
                             }
                         }
 
-                        // Check for bound operations and ensure related types re also selected.
+                        // Check for bound operations and ensure related types are also selected.
                         // In this case related types means return types and parameter types.
                         if (boundOperations.TryGetValue(structuredType, out List<IEdmOperation> operations))
                         {
@@ -200,10 +219,10 @@ namespace Microsoft.OData.ConnectedService.ViewModels
                                     string returnTypeFullName = operation.ReturnType.ToStructuredType()?.FullTypeName() ?? operation.ReturnType.FullName();
                                     AddRelatedType(returnTypeFullName, structuredType.FullTypeName());
 
-                                    if (SchemaTypeModelMap.TryGetValue(returnTypeFullName, out SchemaTypeModel referencedschemaTypeModel)
-                                        && !referencedschemaTypeModel.IsSelected)
+                                    if (SchemaTypeModelMap.TryGetValue(returnTypeFullName, out SchemaTypeModel referencedSchemaTypeModel)
+                                        && !referencedSchemaTypeModel.IsSelected)
                                     {
-                                        referencedschemaTypeModel.IsSelected = true;
+                                        referencedSchemaTypeModel.IsSelected = true;
                                     }
                                 }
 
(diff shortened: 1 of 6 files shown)


unchase:
@paulodero now ready for review.

KenitoInc:
@unchase Rebase your branch

unchase:
> @unchase Rebase your branch

@KenitoInc in the process of adding tests...

unchase:
> Could you add tests for this?

Done!

src/ViewModels/SchemaTypesViewModel.cs
```diff
@@ -74,13 +77,29 @@ public override async Task<PageNavigationResult> OnPageLeavingAsync(WizardLeavin
...
+                    ExcludedSchemaTypeNames.ToList())).ToList();
+                foreach (var operationImport in this._wizard.OperationImportsViewModel.OperationImports)
+                {
+                    if (operationsToExclude.Any(x => x.Name == operationImport.Name))
+                    {
+                        operationImport.IsSelected = false;
+                    }
+                }
+            }
+
```
habbes, line 95: Looks like there's quite some coupling here between the components. And the fact the edm model is fetched again here maybe is an indication that we should consider some refactoring.

src/ViewModels/SchemaTypesViewModel.cs
```diff
@@ -226,6 +245,21 @@ public override async Task<PageNavigationResult> OnPageLeavingAsync(WizardLeavin
...
+                            foreach (var relatedType in relatedTypes)
+                            {
+                                if (SchemaTypeModelMap.TryGetValue(relatedType, out SchemaTypeModel relatedSchemaTypeModel)
+                                    && relatedSchemaTypeModel.IsSelected)
+                                {
+                                    relatedSchemaTypeModel.IsSelected = false;
+                                }
+                            }
+                        }
+                    }
```
habbes, line 262: If related types that were selected caused this schema type to get selected as well in the previous blocks, then how is it that we end up here with the schema type model not being selected?

src/ViewModels/SchemaTypesViewModel.cs
habbes: I don't think this should be a blocker though, if everything else works correctly.

habbes:
@unchase thanks for adding all those extra test cases, I didn't expect you to add test cases besides the one that test the fix you added. Fantastic!

habbes:
@unchase I have tested your PR locally and the bug seems to be fixed. But I have not yet understood where the bug was just by reviewing the code. Could you briefly explain where the bug was and how your code changes fixed it?

src/ViewModels/SchemaTypesViewModel.cs
unchase: I had to add some coupling here to get operations that should be excluded (related with deselected schema types), and, i am agree, we can consider some refactoring. But this can be done in the separate PR.

unchase:
> @unchase I have tested your PR locally and the bug seems to be fixed. But I have not yet understood where the bug was just by reviewing the code. Could you briefly explain where the bug was and how your code changes fixed it?

The bug was fixed in `OnPageLeavingAsync`: when we leaving the page we should to exclude all operations that require the deselected types. It happens in this block of code:
```csharp
if (this._wizard != null)
{
    IEdmModel model = EdmHelper.GetEdmModelFromFile(this._wizard.ConfigODataEndpointViewModel.MetadataTempPath);
    IEnumerable<IEdmOperationImport> operations = EdmHelper.GetOperationImports(model);
    IEnumerable<IEdmOperationImport> operationsToExclude = operations.Where(x => !this._wizard.OperationImportsViewModel.IsOperationImportIncluded(x,
        ExcludedSchemaTypeNames.ToList())).ToList();
    foreach (var operationImport in this._wizard.OperationImportsViewModel.OperationImports)
    {
        if (operationsToExclude.Any(x => x.Name == operationImport.Name))
        {
            operationImport.IsSelected = false;
        }
    }
}
```

src/ViewModels/SchemaTypesViewModel.cs
unchase: Here when we deselect the schema type we just deselect all related types.

unchase: And this block of code is in `schemaTypeModel.PropertyChanged += (s, args) =>` so when we deselect the schema type all its related types should be deselected.

src/ViewModels/SchemaTypesViewModel.cs
unchase: @habbes Refactored

habbes:
@unchase thanks for the clarification.

The feature was designed such that if you left the schema types page with some unselected types while there were still selected types and operations that depended on those unselected types, then you would the required unselected types would be automatically selected and you would get a message prompt informing you of this. That's what [this block of code] is supposed to implement.

But now after looking at it, I see that it only catered to related types and not operation imports. Now I get your solution, thanks.

unchase:
> You've also updated the PropertyChanged event handler such that if a type is deselected, all dependent types are deselected as well. We actually considered implementing it this way at the beginning we decided not to. The concern is that, if you are working with a large schema, with many entities, then having to loop through related items each time could cause lags and affect the user experience. We want to test it extensively with large models before we introduce that feature because it would be harder to remove that feature later on than it would be to introduce it later on. So for now, I propose that you remove that part. When the user leaves the page, the related types will still be verified and auto-selected if required anyway so the user is still safe from generating code that fails to compile.

@habbes I think, that actually for the user, auto-selecting of the related types with showing error message with count of included types when he deselect the specific type is not clear.

For large schemas, the user does not control selected schema types (like in your case) as if they were deselected automatically (like in my case). In this sense, there is no difference.

In addition, the user can deselect types for a long time (for large schema), and after that the whole result of his activity will simply be reset with error message - it's bad isea. He will have to deselect the types again, and he will not know which specific related types caused the error.

habbes:
@unchase 

> In addition, the user can deselect types for a long time (for large schema), and after that the whole result of his activity will simply be reset with error message - it's bad isea. He will have to deselect the types again, and he will not know which specific related types caused the error.

I don't think the user will have to deselect the types again. If they deselect types that are required the generated code would not compile.

Anyway, I have tested this with a model with relatively many types (https://graph.microsoft.com/beta/$metadata) and there wasn't a noticeable lag when deselecting on my machine. I just realized we are caching dependencies upfront so, as you said, the penalty for deselecting is not any worse than for selecting. So we can go ahead and adopt this approach since it provides better UX.

Issue #59, opened 2020-03-23
Improve test coverage


habbes:
## Overall testing gaps:
- Verifying that options set by the user in the wizard are correctly propagated to the CodeGenerator instance
- Verifying that options set in the wizard are saved to the designer data (ConnectedService.json) as expected
- Verifying that settings are loaded from the designer data and used to populate the view when the connected service is being update
- Verify that settings that should be disabled during an update are disabled and settings that are mutually exclusive behave correctly (e.g. settings which should be disabled if T4 templates option is enabled)
- Verify that the T4 templates are generated correctly with the user-defined options stored in the Configuration class as expected
- Verify that settings that should only be available for v4 are treated as expected

### ODataConnectedServiceProvider
This is the entry point of the connected service. It doesn't do much besides `CreateConfiguratorAsync()` which simply creates an `ODataConnectedServiceWizard` instance
and returns it. I don't think it's very critical to test if `ODataConnectedServiceWizard` is thoroughly tested.

### ODataConnectedServiceWizard

This is the actual wizard. A new instance is created when creating or updating a connected service.
Handles the logic for populating the service config from the UI fields (user input) and populating
the UI with previously saved data when updating the connected service. The `GetFinishedServiceInstanceAsync`
is called by VS Connected Service engine and it returns the service instance that contains the config.

This class is lacking in critical tests:
- [x] user settings are loaded when the wizard is created
- [x] When the service is being updated, it should load saved config from `GetExtendedDesignerData()`
and use it to populate the view models accordingly. Test that the right fields are set to the view
models based on the saved config
- [x] Some fields should not be modified after code generation. When the connected service is updating,
ensure that these fields are disabled (e.g. T4 templates, Service Name, Service Endpoint, Generate multiple files, etc.)
- [x] Some fields should only be saved to config (and loaded from saved config) when connected to a V4 endpoint (e.g.
open generated fields in IDE, T4 templates, generate multiple files, etc.). I think this should apply to most
new settings that impact the generated code since we're not adding new features to generated code for v3 endpoints.
- [x] `CreateServiceConfiguration()` returns a `ServiceConfiguration` instance with values corresponding to fields
in the view models.
- [x] `GetFinishedServiceInstanceAsync()` saves user settings, creates a service instance based on the view models values
and service config, and returns that. The result of this method will be passed to the `ODataConnectedServiceHandler` so
testing this is critical in ensuring that values will be correctly propagated to the handler
[...]

PR #118, opened 2020-04-30
Add CodeGenDescriptor tests

habbes:
Addresses #59 

### V4CodeGenDescriptor
This class handles code generation for v4 endpoints. I added test cases to verify that it passes settings to
the `T4CodeGenerator` instance, and that saves the generated code at the right location, but it still lacks
many essential tests:
- [x] `AddNugetPackagesAsync()` should add the relevant OData client libraries to the project (`Microsoft.OData.{Client, Core, Edm}` and `System.Spatial`)
- [x] `AddGeneratedClientCodeAsync()` should add the t4 template files to the project if the corresponding option was enabled, otherwise it should generated c# files and add them to the project
- [x] `AddT4FileAsync()` should add the T4 template .tt and .ttinclude files to the reference folder (`<Project Root>\<Connected Services>\<Service Name>`)
- [x] Configuration values in the .tt file should be replaced with corresponding values from the Service Config. This should account
for service config value which may be `null`
- [x] `AddGeneratedCSharpCodeAsync()` should pass the values from the service config to the code generator instance. This should also account for `null`. This is partially tested but new config properties like `CustomHttpHeaders` are not tested. Should be upated to ensure
all config properties are tested.
- [x] The generated files should should saved at the right location (`<Project Root>\<Connected Services Root>\<Service Name>\<FileName>.cs`). I think there's a test case for this. But it does not handle the case where multiple files are generated.
- [ ] The metadata should be saved in Csdl.xml file in the connected service folder, and should configured as an embedded resource and its path should be used as the temp metadata path of the code generator. *Since this feature is going to be changed soon, we defer adding this test for the feature update*
- [x] for VB projects, we should also test that the code generator is configured accordingly, and generated files are stored with the expected extension

### V3CodeGenDescriptor
- [x] `AddNugetPackageAsync` should install the V3 specific nuget packages
- [x] `AddGeneratedClientCodeAsync` should generate client code in the project language (C# or VB)


ElizabethOkerio:
@habbes  rebase and fix conflicts

habbes:
@ElizabethOkerio rebased and fixed conflicts.

PR #121, opened 2020-05-11
Fix exception when there are multiple cdsl files

### LINKED ISSUE #111: An exception occurs in CreateXmlReader() when there are more than one Csdl files in the assembly
### Describe the bug
When the container tries to read the path of Csdl.xml in CreateXmlReader(), an error occurs when there are more than one Csdl.xml files in the assembly.

private const string filePath = @"Csdl.xml";

```
            {
                try
                {
                    var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
                    **var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));**


//Rest of the code
            }
```

The issue is it tries to find only one ending with Csdl.xml when there are more than one Csdl.xml files. In my case I have generated stubs for 3 different endpoints in 3 different folders. All of them had Csdl.xml

Suggested Fix - Rename the Csdl.xml to **[ServiceName]Csdl.xml**
 
ODataConnectedService v0.9.0
VisualStudio Version v16.5
Microsoft.Odata.Client v7.6.4

#### To Reproduce
1. Generate Stubs for two (or more)different end points.
2. Create an instance of any container

#### Expected behavior
Container instantiated without any error 

#### Actual behavior
An error occurs - System.InvalidOperationException: Sequence contains more than one matching element

#### Additional context
Error Stack
Exception: The type initializer for 'GeneratedEdmModel' threw an exception. ---> System.TypeInitializationException: The type initializer for 'GeneratedEdmModel' threw an exception. ---> System.InvalidOperationException: Sequence contains more than one matching element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
   at ProjectService.DefaultContainer.GeneratedEdmModel.CreateXmlReader() in ... <rest of the stack>

KenitoInc:
Fix #111 
When you add multiple connected services in one project, all csdl files are named `Cdsl.xml`
This causes an exception when reading the embedded resource file since we were trying to read `Csdl.xml` file yet there are multiple of them.

Solution
We are renaming `Csdl.xml` to `[ServiceName]Csdl.xml` to make them unique.

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -84,7 +84,10 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
 
             tempFile = Path.GetTempFileName();
 
-            var metadataFile = Path.Combine(referenceFolder, Common.Constants.CsdlFileName);
+            // Csdl file name is this format [ServiceName]Csdl.xml
+            var csdlFileName = string.Concat(ServiceConfiguration.ServiceName, Common.Constants.CsdlFileNameSuffix);
+            var metadataFile = Path.Combine(referenceFolder, csdlFileName);
+
             await this.Context.HandlerHelper.AddFileAsync(tempFile, metadataFile, new AddFileOptions() { SuppressOverwritePrompt = true });
 
             // Hack!
@@ -93,7 +96,7 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
             var dte = VisualStudio.Shell.Package.GetGlobalService(typeof(EnvDTE.DTE)) as EnvDTE.DTE;
             if (dte != null)
             {
-                var projectItem = this.GetCsdlFileProjectItem(Common.Constants.CsdlFileName);
+                var projectItem = this.GetCsdlFileProjectItem(csdlFileName);
                 projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
             }
 
@@ -122,7 +125,7 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
                 var customHeaders = ServiceConfiguration.CustomHttpHeaders ?? "";
                 text = Regex.Replace(text, "(public const string CustomHttpHeaders = )\"\";", "$1@\"" + customHeaders + "\";");
                 text = Regex.Replace(text, "(public const string MetadataFilePath = )\"\";", "$1@\"" + metadataFile + "\";");
-                text = Regex.Replace(text, "(public const string MetadataFileRelativePath = )\"\";", "$1@\"" + Common.Constants.CsdlFileName + "\";");
+                text = Regex.Replace(text, "(public const string MetadataFileRelativePath = )\"\";", "$1@\"" + csdlFileName + "\";");
                 await writer.WriteAsync(text);
                 await writer.FlushAsync();
             }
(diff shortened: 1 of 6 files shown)


habbes:
@KenitoInc is this update backwards compatible? Will it break clients which were generated using previous versions using the  `Csdl.xml` file name?

paulodero:
> @KenitoInc is this update backwards compatible? Will it break clients which were generated using previous versions using the `Csdl.xml` file name?

I don't think it would cause any problems with current users since the file name only changes when the user updates the client, in which case the code that reads this file is also updated accordingly.

habbes:
@paulodero thanks, that makes sense.

KenitoInc:
> @KenitoInc did you remove Common.Constants.CsdlFileName field in the common constants file? Is it still being used?

Nope, it's still being used

gathogojr:
> > @KenitoInc is this update backwards compatible? Will it break clients which were generated using previous versions using the `Csdl.xml` file name?
> 
> @habbes I don't think it would cause any problems with current users since the file name only changes when the user updates the client, in which case the code that reads this file is also updated accordingly.

@KenitoInc 
When a user updates a Connected Service, will they be stuck with the existing Csdl.xml plus a new metadata file named [Service Name]Csdl.xml? Would it make more sense to search for assembly manifest for qualified metadata file? E.g. `.EndsWith([Service Name]\Csdl.xml)`

marabooy:
> > > @KenitoInc is this update backwards compatible? Will it break clients which were generated using previous versions using the `Csdl.xml` file name?
> > 
> > 
> > @habbes I don't think it would cause any problems with current users since the file name only changes when the user updates the client, in which case the code that reads this file is also updated accordingly.
> 
> @KenitoInc
> When a user updates a Connected Service, will they be stuck with the existing Csdl.xml plus a new metadata file named [Service Name]Csdl.xml? Would it make more sense to search for assembly manifest for qualified metadata file? E.g. `.EndsWith([Service Name].Csdl.xml)`

Probably also should save the file in the service config so that when one is updating the client it checks the old service file name and overwrites it.

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -84,7 +84,10 @@ private async Task AddT4FileAsync()
 
             tempFile = Path.GetTempFileName();
 
-            var metadataFile = Path.Combine(referenceFolder, Common.Constants.CsdlFileName);
+            // Csdl file name is this format [ServiceName]Csdl.xml
+            var csdlFileName = String.Concat(ServiceConfiguration.ServiceName, Common.Constants.CsdlFileName);
```
habbes, line 88: Why not `string.Concat` (using the c# `string` keyword)

habbes: Maybe `Common.Constants.CsdlFileName` should now be called `Common.Constants.CsdlFileNameSuffix` since it no longer represents the full filename, but only the end part.

habbes:
> > > > @KenitoInc is this update backwards compatible? Will it break clients which were generated using previous versions using the `Csdl.xml` file name?
> > > 
> > > 
> > > @habbes I don't think it would cause any problems with current users since the file name only changes when the user updates the client, in which case the code that reads this file is also updated accordingly.
> > 
> > 
> > @KenitoInc
> > When a user updates a Connected Service, will they be stuck with the existing Csdl.xml plus a new metadata file named [Service Name]Csdl.xml? Would it make more sense to search for assembly manifest for qualified metadata file? E.g. `.EndsWith([Service Name].Csdl.xml)`
> 
> Probably also should save the file in the service config so that when one is updating the client it checks the old service file name and overwrites it.

@marabooy but since previous versions did not save the filename in the config, then when they update their client using the new version, it won't know what filename was previously saved and therefore won't know what to overwrite.  Or are you proposing the filename to the config now in case we face this issue in the future?

src/CodeGeneration/V4CodeGenDescriptor.cs

bwn-z:
wanted to let you know that this issue still persists because serviceName is a constant. Maybe it makes sense to give the user the right to reassign through options? --service-name | -sn.

https://github.com/OData/ODataConnectedService/blob/11df9341833a1ba5b7dda8ae105ee49c5558cadf/src/Microsoft.OData.Cli/GenerateCommand.cs#L214

PR #125, opened 2020-05-14
Add EdmHelper Tests

KenitoInc:
Added the tests below:
- [x] GetEdmFromFile reads and returns an edm model from the given file
- [x] GetOperationImports returns operation imports from the model
- [x] GetBoundOperations returns bound operations from the model
- [x] GetSchemaTypes returns all schema types (entity, enum, complex, etc.) from the model
- [x] GetTypeNameFromFullName extract the type name without the namespace from a fully qualified name

### DIFF
--- a/test/ODataConnectedService.Tests/CodeGeneration/CodeGenDescriptorTest.cs
+++ b/test/ODataConnectedService.Tests/CodeGeneration/CodeGenDescriptorTest.cs
@@ -18,14 +18,7 @@ using System.Collections.Generic;
 using System.Net;
 using ODataConnectedService.Tests;
 using System.Text;
-using Microsoft.VisualStudio.ComponentModelHost;
-using System.ComponentModel.Composition.Primitives;
-using System.ComponentModel.Composition.Hosting;
-using System.ComponentModel.Composition;
-using NuGet.VisualStudio;
-using NuGet;
 using System;
-using System.Reflection;
 using System.Linq;
 using ODataConnectedService.Tests.TestHelpers;
 
--- a/test/ODataConnectedService.Tests/Common/EdmHelperTests.cs
+++ b/test/ODataConnectedService.Tests/Common/EdmHelperTests.cs
@@ -1,0 +1,134 @@
+﻿//-----------------------------------------------------------------------------------
+// <copyright file="EdmHelperTests.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//-----------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.OData.ConnectedService.Common;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.OData.Edm;
+
+namespace ODataConnectedService.Tests
+{
+    [TestClass]
+    public class EdmHelperTests
+    {
+        readonly string MetadataPath = Path.GetFullPath("TestMetadataCsdl.xml");
+        [TestMethod]
+        public void TestGetEdmFromFile()
+        {
+            var model = EdmHelper.GetEdmModelFromFile(MetadataPath);
+            Assert.IsNotNull(model);
+        }
+
+        [TestMethod]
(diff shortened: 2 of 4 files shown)


test/ODataConnectedService.Tests/EdmHelperTests.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            var expectedOperationImports = new List<string>()
+            {
+                "GetNearestAirport",
+                "GetPersonWithMostFriends",
+                "ResetDataSource",
+            };
+
+            // Sort the Lists in ascending order
+            expectedOperationImports.Sort();
```
marabooy, line 98: Seems like the asserts are the same maybe you could refactor this to avoid having similar code

test/ODataConnectedService.Tests/EdmHelperTests.cs
```diff
@@ -0,0 +1,135 @@
...
+using Microsoft.OData.ConnectedService.Models;
+using Microsoft.OData.ConnectedService.Common;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.OData.Edm;
+
+namespace ODataConnectedService.Tests
+{
+    [TestClass]
+    public class EdmHelperTests
+    {
```
paulodero, line 19: Any reason why this file `EdmHelperTests.cs` is in the root directory and not in `TestHelpers `folder?

test/ODataConnectedService.Tests/EdmHelperTests.cs
KenitoInc: @marabooy I am not getting you? The tests are different, look at how we are populating the Lists

test/ODataConnectedService.Tests/EdmHelperTests.cs
```diff
@@ -0,0 +1,135 @@
+﻿// Copyright (c) Microsoft Corporation.  All rights reserved.
+// Licensed under the MIT License.  See License.txt in the project root for license information.
```
paulodero, line 2: You could change this to a new format like in the other files.

test/ODataConnectedService.Tests/EdmHelperTests.cs
KenitoInc: I think TestHelpers are helpers for other Test classes

test/ODataConnectedService.Tests/EdmHelperTests.cs
```diff
@@ -0,0 +1,135 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OData.ConnectedService.Models;
+using Microsoft.OData.ConnectedService.Common;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.OData.Edm;
```
paulodero: Remove unnecessary `using `. Below are not required are they?

- `using System;`

- ` using System.Linq;`

- `using System.Text;`

- `using System.Threading.Tasks;`

- `using Microsoft.OData.ConnectedService.Models;`

KenitoInc: Done

test/ODataConnectedService.Tests/EdmHelperTests.cs
paulodero: You are right. you can create a new folder for it.

test/ODataConnectedService.Tests/EdmHelperTests.cs
marabooy: but the assertions are the same. You can change the generation part to helper functions

KenitoInc:
> Please fix the issues highlighted.

@paulodero Am on them

test/ODataConnectedService.Tests/EdmHelperTests.cs
```diff
@@ -0,0 +1,134 @@
+﻿//-----------------------------------------------------------------------------------
+// <copyright file="ODataConnectedServiceHandlerTest.cs" company=".NET Foundation">
```
paulodero, line 2: The file name is not `ODataConnectedServiceHandlerTest.cs`

PR #129, opened 2020-05-18
Add Project Helper tests

KenitoInc:
Adding tests for ProjectHelpers for working with the VS project. 
- [x] `GetProjectFromHierarchy` returns a Project from the given vs hierarchy
- [x] `GetFullPath` returns the full path of the project
- [x] `GetLanguageOption` determine whether the given project is a C# or a VB project

### DIFF
--- a/test/ODataConnectedService.Tests/Common/ProjectHelperTests.cs
+++ b/test/ODataConnectedService.Tests/Common/ProjectHelperTests.cs
@@ -1,0 +1,119 @@
+﻿//-----------------------------------------------------------------------------------
+// <copyright file="ProjectHelperTests.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//-----------------------------------------------------------------------------------
+
+using EnvDTE;
+using Microsoft.OData.ConnectedService;
+using Microsoft.OData.ConnectedService.Common;
+using Microsoft.OData.ConnectedService.Models;
+using Microsoft.OData.ConnectedService.Tests.TestHelpers;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using System;
+using System.Data.Services.Design;
+
+namespace ODataConnectedService.Tests
+{
+    [TestClass]
+    public class ProjectHelperTests
+    {
+        [TestMethod]
+        public void TestGetProjectFromHierarchy()
+        {
+            var serviceConfig = new ServiceConfiguration()
+            {
+                EdmxVersion = new Version(4, 0, 0, 0),
+                ServiceName = "TestService",
+                UseDataServiceCollection = false,
+                MakeTypesInternal = true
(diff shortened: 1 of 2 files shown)


test/ODataConnectedService.Tests/ProjectHelperTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                EdmxVersion = new Version(4, 0, 0, 0),
+                ServiceName = "TestService",
+                UseDataServiceCollection = false,
+                MakeTypesInternal = true
+            };
+            var context = SetupContext(serviceConfig);
+            var project = ProjectHelper.GetProjectFromHierarchy(context.ProjectHierarchy);
+            var languageOption = ProjectHelper.GetLanguageOption(project);
+            Assert.IsNotNull(languageOption);
+            Assert.AreEqual(LanguageOption.GenerateCSharpCode, languageOption);
```
habbes, line 66: I think you should also add a test for VB

test/ODataConnectedService.Tests/ProjectHelperTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            var serviceConfig = new ServiceConfiguration()
+            {
+                EdmxVersion = new Version(4, 0, 0, 0),
+                ServiceName = "TestService",
+                UseDataServiceCollection = false,
+                MakeTypesInternal = true
+            };
+            var context = SetupContext(serviceConfig);
+            var project = ProjectHelper.GetProjectFromHierarchy(context.ProjectHierarchy);
+            Assert.IsNotNull(project);
```
habbes, line 32: Could you also test some other property of the `project`, just to ensure that it's the expected project and not just a arbitrary non-null object?

test/ODataConnectedService.Tests/ProjectHelperTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                EdmxVersion = new Version(4, 0, 0, 0),
+                ServiceName = "TestService",
+                UseDataServiceCollection = false,
+                MakeTypesInternal = true
+            };
+            var context = SetupContext(serviceConfig);
+            var project = ProjectHelper.GetProjectFromHierarchy(context.ProjectHierarchy);
+            var fullpath = ProjectHelper.GetFullPath(project);
+            Assert.IsNotNull(fullpath);
```
marabooy, line 48: Looks redundant to me as this is a primitive. The Assert equals should suffice for this check

test/ODataConnectedService.Tests/ProjectHelperTests.cs
KenitoInc: @habbes Done

test/ODataConnectedService.Tests/ProjectHelperTests.cs
KenitoInc: Done

test/ODataConnectedService.Tests/ProjectHelperTests.cs
KenitoInc: @marabooy Fixed

PR #131, opened 2020-05-18
Fix #130 : Deselecting a schema type does not exclude its bound operations

### LINKED ISSUE #130: Deselecting a schema type does not exclude its bound operations
### Describe the bug
Deselecting a schema type does not exclude its bound operations. This operations will be added to generated code. Will get compilation errors.

Microsoft.OData.ConnectedService v0.10.x

#### To Reproduce

1. Create an OData Connected Service based on the [Microsoft Graph service](https://graph.microsoft.com/v1.0/$metadata);
2. Move to the schema types page, all types should be selected by default;
3. Deselect, for example, `user` schema type (which has bound operations);
4. Proceed to the function/action imports page, there is no operation imports;
5. Click **Finish** to generate code;
6. Try to build project.

#### Expected behavior
Deselecting schema types should deselect its bound operations.

#### Actual behavior
Deselecting schema types does not affect its bound operations.

#### Screenshots
* Use [Microsoft Graph service](https://graph.microsoft.com/v1.0/$metadata) metadata:

page1

* Deselecting `user` schema type:

page2

* `user` schema type and one of its bound operation:

edm

* There is no operation imports:

page3

* Compilations errors:

error

unchase:
Fix #130

This adds bound operations to schema types in the **"Schema Types Selection"** wizard page where the user can select which bound operations are included in the generated code:

bound_operations_feature3

boundOperations

Features:
* Display schema types with its bound operations in the treeview
* Add buttons for select/deselect all bound operations for all schema types
* Add context menu to select/deselect all bound operations for specified schema types
* Display selected/all schema types count and bound operations count
* Fix a few bugs

Tested for [Microsoft Graph service](https://graph.microsoft.com/v1.0/$metadata) and [Trippin sample service](https://services.odata.org/TripPinRESTierService/$metadata).

----

Added the tests below:

To **SchemaTypesViewModelTests**:
- [x] `ExcludeSchemaTypes_ShouldDeselectTheSpecifiedTypesWithSpecifiedBoundOperations`;
- [x] `SelectBoundOperation_ShouldSelectItsRequiredType`;
- [x] `DeselectSchemaType_ShouldDeselectItsBoundOperations`;
- [x] `ExcludeBoundOperations_ShouldDeselectTheSpecifiedBoundOperations`;
- [x] `ExcludedBoundOperationsNames_ShouldReturnNamesOfDeselectedBoundOperationsOrderedByName`;
- [x] `ClearBoundOperationList_ShouldResetBoundOperationsForTheSpecificTypeToAnEmptyList`;
- [x] `SelectAllBoundOperations_ShouldMarkAllBoundOperationsForAllTypesAsSelected`;
- [x] `SelectAllBoundOperationsForSchemaType_ShouldMarkAllBoundOperationsForTheSpecificTypeAsSelected`;
- [x] `DeselectAllBoundOperations_ShouldMarkBoundOperationsForAllTypesAsNotSelected`;
- [x] `DeselectAllBoundOperationsForSchemaType_ShouldMarkBoundOperationsForTheSpecificTypeAsNotSelected`.

To **SchemaTypeModelToVisibilityConverterTests**:
- [x] `Convert_ShouldReturnVisible_ForSchemaTypeWithBoundOperations`;
- [x] `Convert_ShouldReturnCollapsed_ForSchemaTypeWithoutBoundOperations`;
- [x] `Convert_ShouldThrowNotSupportedException_ForNonVisibilityTargetType`;
- [x] `ConvertBack_ShouldThrowNotSupportedException`.

To **CodeGenDescriptorTest**:
- [x] `TestAddGeneratedClientCode_GeneratesT4TemplateFiles_WithIncludeT4File_WithExcludedSchemaTypes`;
- [x] `TestAddGeneratedClientCode_GeneratesT4TemplateFiles_WithIncludeT4File_WithExcludedOperationImports`;
- [x] `TestAddGeneratedClientCode_GeneratesT4TemplateFiles_WithIncludeT4File_WithExcludedBoundOperations`.


unchase:

habbes:
@unchase  when you select a bound operation, does it also auto-select related types (i.e. the type its bound + that types dependencies, the parameter and return types and their dependencies)?

unchase:
> @unchase when you select a bound operation, does it also auto-select related types (i.e. the type its bound + that types dependencies, the parameter and return types and their dependencies)?

@habbes yes, all related types auto-select/deselect in `boundOperationModel.PropertyChanged += (s, args) =>`

src/Models/BoundOperationModel.cs
```diff
@@ -0,0 +1,30 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="OperationImportModel.cs" company=".NET Foundation">
```
paulodero, line 2: Change file name

src/ViewModels/BoundOperationsViewModel.cs
```diff
@@ -0,0 +1,235 @@
+﻿//---------------------------------------------------------------------------------
+// <copyright file="OperationImportsViewModel.cs" company=".NET Foundation">
```
paulodero, line 2: change file name

paulodero:
@unchase the build pipeline is now public. You should be able to check for the build status.

unchase:
> @unchase the build pipeline is now public. You should be able to check for the build status.

Yes it is.

src/Models/BoundOperationModel.cs
unchase: Done.

src/ViewModels/BoundOperationsViewModel.cs
unchase: Done.

unchase:
> A few issues to be fixed.

@paulodero done!

unchase:
> @unchase Is it possible to list the bound operations under the related types in a tree-like structure? I think it would make a better visual impression than to list them in a separate tab, especially when we have very many bound operations like the example you have provided.

@paulodero The deselection of types automatically deselect all related types in such a way that when leaving the page, an error with simultaneous including required types cannot occur.

Deselection and selection of types are consistent with related types in both directions. In addition, it consistent with deselection and selection of bound operations.

A situation cannot arise when types or bound operations are selected without its required types.

unchase:
> @unchase Is it possible to list the bound operations under the related types in a tree-like structure? I think it would make a better visual impression than to list them in a separate tab, especially when we have very many bound operations like the example you have provided.

@paulodero done!

src/Converters/SchemaTypeModelToVisibilityConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class SchemaTypeModelToVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is SchemaTypeModel schemaTypeModel && schemaTypeModel.BoundOperations.Any())
+                return Visibility.Visible;
+
+            return Visibility.Collapsed;
+        }
```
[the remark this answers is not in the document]
unchase: @paulodero done!

unchase:
> This is a nice feature @unchase. It mostly worked as expected, but I faced the following issues:
> 
> * When you update a service that had excluded bound operations, all bound operations are still displayed as selected. You can see in the `ConnectedService.json` file that the excluded operations were correctly saved in the `ExcludedBoundOperations` list, but it appears that this is being ignored when updating the service.
> * When you generate T4 templates, the `ExcludedBoundOperations` config variable should be populate with a comma-separated list of the excluded operations, similar to how other excluded types are handled, but instead the generated as an empty string. When you populate the list manually, it does work as expected. But it should be auto-populated by the wizard.
> * Since you're displaying the bound operations in a nested tree view (which I think is a great idea), I think it's no longer necessary to display the full  name of the type being bound to as a parameter (since it's implied by the tree-nesting). This is just an opinion.
> * Please also add relevant tests to verify the points mentioned here.
> * I don't know if you already have, but in case you haven't should also add test to ensure proper of the property to and from string when using the code generator is used on the command line

@habbes All fixes based on comments have been added. Please check it.

PR #134, opened 2020-05-28
Fix #133 : Generated client uses wrong capitalization

### LINKED ISSUE #133: Generated client uses wrong capitalization for 'False'
When I generate my OData Client, I get the following code:

```
private bool _IsDisplayed = False;
partial void OnIsDisplayedChanging(bool value);
partial void OnIsDisplayedChanged();
```

'**F**alse' is incorrect in C#, it should be '**f**alse' (lowercase 'f'). It's a little annoying, but ultimately something I can fix after re-generating the client every time.

I'm not sure if it matters, but the constructor in my API for the class does explicitly set `IsDisplayed = false`.

Using Microsoft.OData.ConnectedService for generating, using Microsoft.AspNet.OData 7.4.0 on my API.

unchase:
Fix #133

Added the tests below:

To **ODataT4CodeGeneratorTests:**

- [x] `CodeGenEntityBooleanPropertyWithDefaultValueEdmx`.


unchase:

unchase:
@marabooy rebased.

ElizabethOkerio:
@unchase rebase.

PR #177, opened 2020-11-13
Fix #48 Metadata URL with values in Query String

### LINKED ISSUE #48: Metadata URL with values in Query String
Hi.

The url to get the metadata from my OData API requires setting the Api version in a parameter sent in the url query string.

http://[host]:[port]/odata/$metadata **?api-version=1.0**

In this case, when you try to configure the connected service, an error occurs because the extension tries to concatenate "$metadata" to the end of the URL.

How can I use this feature with the exposed scenario?

Thank you

Victor Perez

image

facundo91:
Fix #48 Metadata URL with values in Query String

* Fix applied
* Test added
* Some typos fixed

### DIFF
--- a/src/ViewModels/ConfigODataEndpointViewModel.cs
+++ b/src/ViewModels/ConfigODataEndpointViewModel.cs
@@ -101,7 +101,7 @@ namespace Microsoft.OData.ConnectedService.ViewModels
             if (this.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || this.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!this.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!this.Endpoint.Contains("$metadata"))
                 {
                     this.Endpoint = this.Endpoint.TrimEnd('/') + "/$metadata";
                 }
--- a/test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
+++ b/test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
@@ -54,11 +54,25 @@ namespace ODataConnectedService.Tests.ViewModels
             Assert.IsFalse(pageNavigationResult.IsSuccess);
             Assert.IsTrue(pageNavigationResult.ShowMessageBoxOnFailure);
 
+            //Provide a url with $metadata somewhere in the url
+            configOdataEndPointViewModel.Endpoint = "http://mysite/ODataService/$metadata?api-version=2.0";
+            pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            //Check if the url is detected as valid and kept without no modification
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=177) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=177) before we can accept your contribution.<br/><br/>:x: facundo91 [sign now](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=177)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OData/ODataConnectedService?pullRequest=177) it.</sub>

marabooy:
@facundo91 can you rebase this kindly.

marabooy (approved):

src/ViewModels/ConfigODataEndpointViewModel.cs
```diff
@@ -101,7 +101,7 @@ internal string GetMetadata(out Version edmxVersion)
             if (this.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || this.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!this.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!this.Endpoint.Contains("$metadata"))
                 {
                     this.Endpoint = this.Endpoint.TrimEnd('/') + "/$metadata";
```
gathogojr, line 106: @facundo91 @marabooy Do we need to make any change to this line or are we happy losing the query string part?

marabooy: I think this should change as the current logic won't work for valid uri's and may generate some invalid cases.
@facundo91 can you fix this in this Pr so that the query string is preserved even when we modify the URI.

marabooy, line 106: @facundo91 Kindly address the above comment.

marabooy (changes requested):

facundo91:
I will address this PR today or the day after. Thank you

src/ViewModels/ConfigODataEndpointViewModel.cs
facundo91: This is the requested in the Issue #48, am I loosing something?
This should work with urls as:  `http://[host]:[port]/odata/$metadata?api-version=1.0`
It passes all existing tests, I could add a new test case or test scenario if missing.

marabooy: So the issue in this case is that a url that has the following data will probably not work 
`http://host/odata/?api-version=2` the above will remove the query string.

marabooy: You could in this case just save the query string and then append it at the end or use a url builder to break it up and add it back.

KenitoInc:
Hey @facundo91 
Kindly
- Address review comments
- Rebase the branch and resolve conflicts.

Nthemba:
Issue fixed by PR #311

PR #187, opened 2020-12-22
Set the most logical config defaults for the wizard

### LINKED ISSUE #181: Types emitted on OData client should implement INotifyPropertyChanged interface
### Description
OData client uses `DataServiceCollection<T>` to support property tracking.

#### Example:
```csharp
DataServiceCollection<Order> orders = new DataServiceCollection<Order>(dataServiceContext.Orders.Where(d => d.Id == 1));
// Update Amount
orders[0].Amount = 37;
dataServiceContext.SaveChanges();
```
The payload will comprise only of the updated properties and will be sent as a PATCH request:
```
{
    "@odata.type": "#NS.Models.Order",
    "Amount": 37
}
```

However, to use `DataServiceCollection<T>`, the client type needs to implement `INotifyPropertyChanged` interface. The `Order` type for example would look as follows:
```csharp
[Key("Id")]
public partial class Order : BaseEntityType, global::System.ComponentModel.INotifyPropertyChanged
{
	public static Order CreateOrder(int ID, DateTimeOffset orderDate, decimal amount)
	{
		Order order = new Order();
		order.Id = ID;
		order.OrderDate = orderDate;
		order.Amount = amount;
		return order;
	}

	public virtual int Id
	{
		get
		{
			return this._Id;
		}
		set
		{
			this.OnIdChanging(value);
			this._Id = value;
			this.OnIdChanged();
			this.OnPropertyChanged("Id");
		}
	}
	private int _Id;
	partial void OnIdChanging(int value);
	partial void OnIdChanged();

	public virtual DateTimeOffset OrderDate
	{
		get
		{
			return this._OrderDate;
		}
		set
		{
			this.OnOrderDateChanging(value);
			this._OrderDate = value;
			this.OnOrderDateChanged();
			this.OnPropertyChanged("OrderDate");
		}
	}
	private DateTimeOffset _OrderDate;
	partial void OnOrderDateChanging(DateTimeOffset value);
	partial void OnOrderDateChanged();

	public virtual decimal Amount
	{
		get
		{
			return this._Amount;
		}
		set
		{
			this.OnAmountChanging(value);
			this._Amount = value;
			this.OnAmountChanged();
			this.OnPropertyChanged("Amount");
		}
	}
	private decimal _Amount;
	partial void OnAmountChanging(decimal value);
	partial void OnAmountChanged();
	/// <summary>
	/// This event is raised when the value of the property is changed
	/// </summary
	public event global::System.ComponentModel.PropertyChangedEventHandler PropertyChanged;
	/// <summary>
	/// The value of the property is changed
	/// </summary>
	/// <param name="propertyName">Property name</param>
	protected virtual void OnPropertyChanged(string propertyName)
	{
		if (this.PropertyChanged != null)
		{
			this.PropertyChanged(this, new global::System.ComponentModel.PropertyChangedEventArgs(propertyName));
		}
	}
}
```
**Note:** The class has been stripped to make it less verbose with emphasis being placed on the fragments that support property tracking. Each property setter calls the `OnPropertyChanged(string)` method.

Currently, the client types emitted by OData Connected Service extension do not implement the `INotifyPropertyChanged` interface. An exception is thrown if one tries to use `DataServiceCollection<T>` in OData client.

> System.InvalidOperationException: 'An attempt to track an entity or complex type failed because the entity or complex type 'NS.Models.Order' does not implement the INotifyPropertyChanged interface.'

This in turn makes it impossible to support property tracking and alternatives available result into the whole entity being sent over the wire instead of only the updated properties.

#### Additional context
The CodeGen tool used to emit client types that implement `INotifyPropertyChanged` interface. It'd seem that when a decision was made to discontinue active development of CodeGen, this capability was not migrated to OData Connected Service.

gathogojr:
This pull request fixes #181 

Set the most logical config defaults for the wizard. Below is a detailed explanation of the changes:

1. Restore OCS wizard default settings that were inadvertently dropped with the introduction of the settings persistence feature, i.e.
   - GeneratedFileNamePrefix => `Reference`
   - NamespacePrefix => `Reference`
   - UseDataServiceCollection => `true`
   - EnableNamingAlias => `true`
   - IgnoreUnexpectedElementsAndAttributes => `true`
   - ServiceName => `OData Service` 

2. ~Fix handling of **most recently used** (MRU) endpoints. Presently, the `UserSettings` class exposes MRU endpoints as an `ObservableCollection` that one is able to manipulate directly (add/remove/etc) and violate the whole idea of "most recently used". This is fixed by changing `MruEndpoints` property type to `ReadOnlyObservableCollection` and providing a public method `AddMruEndpoint` through which the MRU endpoints collection can be manipulated~

3. Fix binding of user settings to the wizard pages. There was a lot of unnecessary duplication of code to support binding with lots of copying logic in multiple places. This is fixed by having all the wizard pages bind directly to the properties in a `UserSettings` instance.

4. Tests would change the settings file that is used when creating Connected Services from Visual Studio. This is fixed by making it possible to create a "named" user settings config file in test scenarios

5. Refactor to make the code leaner and cleaner

### DIFF
--- a/src/Common/ExtensionMethods.cs
+++ b/src/Common/ExtensionMethods.cs
@@ -1,0 +1,46 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="ExtensionMethods.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//----------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.OData.ConnectedService.Common
+{
+    internal static class ExtensionMethods
+    {
+        /// <summary>
+        /// Copies public properties with matching name and type from one object to the other.
+        /// </summary>
+        /// <param name="target">The target.</param>
+        /// <param name="source">The source.</param>
+        public static void CopyPropertiesFrom(this object target, object source)
+        {
+            if (source == null)
+            {
+                return;
+            }
+
+            var fromProperties = source.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                .Where(d => d.GetGetMethod() != null);
+            var toProperties = target.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                .Where(d => d.GetSetMethod() != null);
+
+            foreach(var toProperty in toProperties)
+            {
+                var fromProperty = fromProperties.SingleOrDefault(d => 
+                    d.Name.Equals(toProperty.Name, StringComparison.Ordinal) 
+                    && toProperty.PropertyType.IsAssignableFrom(d.PropertyType));
+
+                if (fromProperty != null)
+                {
+                    toProperty.SetValue(target, fromProperty.GetValue(source));
+                }
+            }
+        }
+    }
+}
--- a/src/Common/UserSettingsPersistenceHelper.cs
+++ b/src/Common/UserSettingsPersistenceHelper.cs
@@ -36,7 +36,8 @@ namespace Microsoft.OData.ConnectedService.Common
                         {
                             // note: this overwrites existing settings file if it exists
                             stream = file.OpenFile(fileName, FileMode.Create);
-                            using (var writer = XmlWriter.Create(stream))
+
+                            using (var writer = XmlWriter.Create(stream, new XmlWriterSettings { Indent = true }))
                             {
                                 var dcs = new DataContractSerializer(userSettings.GetType());
                                 dcs.WriteObject(writer, userSettings);
@@ -112,7 +113,7 @@ namespace Microsoft.OData.ConnectedService.Common
 
         private static string GetStorageFileName(string providerId, string name)
         {
-            return providerId + "_" + name + ".xml";
+            return providerId + "." + name + ".xml";
         }
 
         private static IsolatedStorageFile GetIsolatedStorageFile()
--- a/src/Models/UserSettings.cs
+++ b/src/Models/UserSettings.cs
@@ -5,147 +5,466 @@
 // </copyright>
 //----------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using Microsoft.OData.ConnectedService.Common;
 using Microsoft.VisualStudio.ConnectedServices;
+using Newtonsoft.Json;
 
 namespace Microsoft.OData.ConnectedService.Models
 {
     [DataContract]
-    internal class UserSettings
+    internal class UserSettings : INotifyPropertyChanged
     {
-        private const string Name = "Settings";
+        #region Const Members
+
         private const int MaxMruEntries = 10;
 
-        private ConnectedServiceLogger logger;
+        #endregion Const Members
+
+        #region Readonly Members
+
+        private readonly ConnectedServiceLogger logger;
+
+        private readonly string configName;
+
+        #endregion Readonly Members
+
+        #region Private Members
+
+        private ObservableCollection<string> mruEndpoints;
+
+        private string serviceName;
+
+        private string endpoint;
+
+        private string generatedFileNamePrefix;
+
+        private bool useNamespacePrefix;
+
+        private string namespacePrefix;
+
+        private bool useDataServiceCollection;
+
+        private bool makeTypesInternal;
+
+        private bool openGeneratedFilesInIDE;
+
+        private bool generateMultipleFiles;
+
+        private string customHttpHeaders;
+
+        private bool enableNamingAlias;
+
+        private bool ignoreUnexpectedElementsAndAttributes;
+
+        private bool includeT4File;
+
+        private bool includeWebProxy;
+
+        private string webProxyHost;
+
+        private bool includeWebProxyNetworkCredentials;
+
+        private string webProxyNetworkCredentialsUsername;
+
+        private string webProxyNetworkCredentialsPassword;
+
+        private string webProxyNetworkCredentialsDomain;
+
+        private bool includeCustomHeaders;
+
+        private List<string> excludedOperationImports;
+
+        private List<string> excludedBoundOperations;
+
+        private List<string> excludedSchemaTypes;
+
+        #endregion Private Members
+
+        public event PropertyChangedEventHandler PropertyChanged;
 
         [DataMember]
-        public ObservableCollection<string> MruEndpoints { get; set; }
-
-        [DataMember]
-        public string ServiceName { get; set; }
-
-        [DataMember]
-        public string Endpoint { get; set; }
-
-        [DataMember]
-        public string GeneratedFileNamePrefix { get; set; }
-
-        [DataMember]
-        public bool UseNamespacePrefix { get; set; }
-
-        [DataMember]
-        public string NamespacePrefix { get; set; }
-
-        [DataMember]
-        public bool UseDataServiceCollection { get; set; }
-
-        [DataMember]
-        public bool MakeTypesInternal { get; set; }
-
-        [DataMember]
-        public bool OpenGeneratedFilesInIDE { get; set; }
-
-        [DataMember]
-        public bool GenerateMultipleFiles { get; set; }
-
-        [DataMember]
-        public string CustomHttpHeaders { get; set; }
-
-        [DataMember]
-        public bool EnableNamingAlias { get; set; }
-
-        [DataMember]
-        public bool IgnoreUnexpectedElementsAndAttributes { get; set; }
-
-        [DataMember]
-        public bool IncludeT4File { get; set; }
-
-        [DataMember]
-        public bool IncludeWebProxy { get; set; }
-
-        [DataMember]
-        public string WebProxyHost { get; set; }
-
-        [DataMember]
-        public bool IncludeWebProxyNetworkCredentials { get; set; }
-
-        [DataMember]
-        public string WebProxyNetworkCredentialsUsername { get; set; }
-
-        [DataMember]
-        public string WebProxyNetworkCredentialsPassword { get; set; }
-
-        [DataMember]
-        public string WebProxyNetworkCredentialsDomain { get; set; }
-
-        [DataMember]
-        public bool IncludeCustomHeaders { get; set; }
-
-        [DataMember]
-        public List<string> ExcludedOperationImports { get; set; }
-
-        [DataMember]
-        public List<string> ExcludedBoundOperations { get; set; }
-
-        [DataMember]
-        public List<string> ExcludedSchemaTypes { get; set; }
-
-        public UserSettings()
+        public ObservableCollection<string> MruEndpoints
         {
-            this.MruEndpoints = new ObservableCollection<string>();
+            get { return mruEndpoints; }
+            set
+            {
+                mruEndpoints = value;
+                OnPropertyChanged();
+            }
         }
 
+        [DataMember]
+        public string ServiceName
+        {
+            get { return serviceName; }
+            set
+            {
+                serviceName = value;
+                OnPropertyChanged();
+            }
+        }
+
+        [DataMember]
+        public string Endpoint
+        {
+            get { return endpoint; }
+            set
+            {
+                endpoint = value;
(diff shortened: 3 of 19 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=187) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=187) before we can accept your contribution.<br/><br/>:x: gathogojr [sign now](https://cla.dotnetfoundation.org/OData/ODataConnectedService?pullRequest=187)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OData/ODataConnectedService?pullRequest=187) it.</sub>

src/Common/UserSettingsPersistenceHelper.cs
```diff
@@ -112,7 +113,7 @@ public static void Save(object userSettings, string providerId, string name, Act
 
         private static string GetStorageFileName(string providerId, string name)
         {
-            return providerId + "_" + name + ".xml";
+            return providerId + "." + name + ".xml";
```
habbes, line 116: What was the purpose of changing from `_` to `.`  in the filename?

src/Models/UserSettings.cs
```diff
@@ -16,13 +17,24 @@ namespace Microsoft.OData.ConnectedService.Models
...
-        private ConnectedServiceLogger logger;
+        private readonly ConnectedServiceLogger logger;
 
-        [DataMember]
-        public ObservableCollection<string> MruEndpoints { get; set; }
+        private readonly string configName;
+
+        private Action<UserSettings> onUserSettingsLoaded = (userSettings) =>
+        {
+            userSettings.MruEndpoints = new ReadOnlyObservableCollection<string>(userSettings.mruEndpoints);
```
habbes, line 28: How come here you are not checking whether `userSettings != null` but in other cases you in this file you are?

src/Models/UserSettings.cs
```diff
@@ -93,59 +105,91 @@ internal class UserSettings
...
+                {
+                    // onLoaded action is currently triggered only if settings are loaded from config file
+                    // Do this defensively all the same...
+                    if (userSettings != null)
+                    {
+                        this.mruEndpoints = new ObservableCollection<string>(userSettings.mruEndpoints);
+                        this.MruEndpoints = new ReadOnlyObservableCollection<string>(this.mruEndpoints);
+                    }
+                },
+                logger: logger);
```
habbes, line 137: If I understand correctly, this automatically attempts to load the settings from file when you instantiate `new UserSettings()` even without calling `userSettings.Load()`?

habbes (approved):
Thanks for cleaning things up.

src/Models/UserSettings.cs
```diff
@@ -72,10 +74,10 @@ internal class UserSettings
         [DataMember]
         public bool IncludeWebProxyNetworkCredentials { get; set; }
 
-        [DataMember]
+        [IgnoreDataMember]
```
KenitoInc, line 77: Also add `[IgnoreDataMember]` to `CustomHttpHeaders` since we my have authentication tokens

src/Models/UserSettings.cs
```diff
@@ -93,59 +95,89 @@ internal class UserSettings
...
-            this.MruEndpoints = new ObservableCollection<string>();
+            this.configName = configName;
+            this.logger = logger;
+            // Desired defaults
+            GeneratedFileNamePrefix = Constants.DefaultReferenceFileName;
+            NamespacePrefix = Constants.DefaultReferenceFileName;
+            ServiceName = Constants.DefaultServiceName;
+            UseDataServiceCollection = true; // To support entity and property tracking
+            EnableNamingAlias = true; // To force upper camel case in the event that entities are named in lower camel case
+            IgnoreUnexpectedElementsAndAttributes = true;
```
KenitoInc, line 108: Maybe you add a comment similar to the way you did for `UseDataServiceCollection` and `EnableNamingAlias`

src/Models/UserSettings.cs
```diff
@@ -16,10 +17,11 @@ namespace Microsoft.OData.ConnectedService.Models
...
-        private const string Name = "Settings";
         private const int MaxMruEntries = 10;
 
-        private ConnectedServiceLogger logger;
+        private readonly ConnectedServiceLogger logger;
+
+        private readonly string configName;
 
         [DataMember]
         public ObservableCollection<string> MruEndpoints { get; set; }
```
KenitoInc, line 27: Since we have a public `get` and `set` here, I think `MruEndPoints` can be manipulated wrongly here.

src/ODataConnectedServiceWizard.cs
```diff
@@ -32,90 +32,87 @@ internal class ODataConnectedServiceWizard : ConnectedServiceWizard
...
-            if (this.Context != null && this.Context.IsUpdating)
+
+            if (Context != null && Context.IsUpdating)
             {
-                ConfigODataEndpointViewModel.Endpoint = this._serviceConfig?.Endpoint;
-                ConfigODataEndpointViewModel.EdmxVersion = this._serviceConfig?.EdmxVersion;
-                ConfigODataEndpointViewModel.ServiceName = this._serviceConfig?.ServiceName;
-                ConfigODataEndpointViewModel.CustomHttpHeaders = this._serviceConfig?.CustomHttpHeaders;
+                LoadUserSettingsFromServiceConfig();
+                ConfigODataEndpointViewModel.EdmxVersion = ServiceConfig?.EdmxVersion;
```
KenitoInc, line 70: Why did we stop adding the `Endpoint` and `ServiceName` to the View model?

KenitoInc (changes requested):
A few comments and questions!

src/Common/UserSettingsPersistenceHelper.cs
```diff
@@ -36,7 +36,8 @@ public static void Save(object userSettings, string providerId, string name, Act
                         {
                             // note: this overwrites existing settings file if it exists
                             stream = file.OpenFile(fileName, FileMode.Create);
-                            using (var writer = XmlWriter.Create(stream))
+
+                            using (var writer = XmlWriter.Create(stream, new XmlWriterSettings { Indent = true }))
```
marabooy, line 40: Do we need to indent this or can this remain the same as the file is not usually read by human?

src/Models/UserSettings.cs
```diff
@@ -93,59 +95,92 @@ internal class UserSettings
...
+                        foreach(var endpoint in userSettings.MruEndpoints)
+                        {
+                            this.MruEndpoints.Add(endpoint);
+                        }
+                    }
+                },
+                logger: logger);
         }
 
+        public UserSettings(string configName = "Settings")
```
marabooy, line 131: Docs for the ctor.

src/ODataConnectedServiceWizard.cs
```diff
@@ -32,90 +32,88 @@ internal class ODataConnectedServiceWizard : ConnectedServiceWizard
...
                 // Restore the advanced settings to UI elements.
                 AdvancedSettingsViewModel.PageEntering += AdvancedSettingsViewModel_PageEntering;
             }
 
-            this.Pages.Add(ConfigODataEndpointViewModel);
-            this.Pages.Add(SchemaTypesViewModel);
-            this.Pages.Add(OperationImportsViewModel);
-            this.Pages.Add(AdvancedSettingsViewModel);
-            this.IsFinishEnabled = true;
+            Pages.Add(ConfigODataEndpointViewModel);
```
marabooy, line 81: To this. or not to?

src/Common/ExtensionMethods.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.OData.ConnectedService.Common
+{
+    internal static class ExtensionMethods
+    {
+        public static void CopyPropertiesFrom(this object target, object source)
```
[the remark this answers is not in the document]
gathogojr: Done

src/Common/UserSettingsPersistenceHelper.cs
gathogojr: During development when you may need to verify the content, it's extremely hard to follow. I reasoned there would be no harm in having it properly formatted

src/Common/UserSettingsPersistenceHelper.cs
gathogojr: The new config file gives the users a new clean slate with default settings as they originally used to be. The old config file would for example have `EnableNamingAlias`, `UseDataServiceCollection` and `IgnoreUnexpectedElementsAndAttributes` default values all set with a value of `false` due to previous mishandling of the default values.

src/Models/UserSettings.cs
gathogojr: Done

src/ODataConnectedServiceWizard.cs
gathogojr: The approach I adopting is to have the control binding to a single `UserSettings` instance newed up at [this point](https://github.com/OData/ODataConnectedService/pull/187/files#diff-00999e34e0165406b5257c8c0a958b3a089653a8b8c0cb618d1be46af253f0d3R52). `Endpoint` and `ServiceName` are properties of `UserSettings` and the controls bind to them. That's the reason I'm not adding them to the view model.

src/ODataConnectedServiceWizard.cs
gathogojr: Whatever code analyzer we're using (or VS default analyzer) considers `this.` superfluous so I dropped it in quite a number of places where possible

src/Models/UserSettings.cs
gathogojr: Fixed indirectly after refactoring the logic

src/Models/UserSettings.cs
gathogojr: Yes. We're primarily interested in the MRU endpoints since we always use them to populate the Endpoint dropdown. That way we have everything that we need for purposes of rendering the view. Note that we don't use any other setting from the config file.

src/Models/UserSettings.cs
gathogojr: Done

src/Models/UserSettings.cs
gathogojr: Done

src/Models/UserSettings.cs
gathogojr: That is a concern I also had and set out to address but in the end I had to trade off since using a `ReadOnlyObservableCollection` inhibited two-way binding with the view causing the wizard not to work as expected

src/Models/UserSettings.cs
```diff
@@ -5,147 +5,466 @@
...
+
             if (index >= 0)
             {
-                // Ensure there aren't any duplicates in the list.
-                for (var i = mruList.Count - 1; i > index; i--)
+                // Remove possible duplicates
+                for (var i = this.MruEndpoints.Count - 1; index > 0 && i > index; i--)
                 {
-                    if (EqualityComparer<T>.Default.Equals(mruList[i], item))
+                    if (this.MruEndpoints[i].Equals(mruEndpoint, StringComparison.Ordinal))
```
KenitoInc, line 448: We need to be consistent on the use or otherwise of `this.`

src/Properties/AssemblyInfo.cs
```diff
@@ -39,8 +39,8 @@
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("0.11.1.0")]
-[assembly: AssemblyFileVersion("0.11.1.0")]
+[assembly: AssemblyVersion("0.11.2.0")]
```
KenitoInc, line 42: Was the version bump deliberate?
Keep an eye on it during release

KenitoInc (commented):
Minor comments

src/Properties/AssemblyInfo.cs
gathogojr: @KenitoInc We'll most definitely need to bump up the version but I'll verify the appropriate version number to use

src/Models/UserSettings.cs
gathogojr: Noted and fixed

KenitoInc (approved):
LGTM

Issue #184, opened 2020-12-12
Unable to parse SAP OData v4 Edmx


Tiberriver256:
### Describe the bug
First error was "The attribute 'BaseTerm' was not expected in the given context."

I removed that section from the edmx and then started getting errors about the type `Edm.PropertyPath`

#### Version of the Project affected
ODataConnectedService v0.11.1
VisualStudio Version v16.4
Microsoft.Odata.Client v7.6.x

#### To Reproduce

See the attached edmx. This is part of a standard include for all SAP OData v4 services so, unfortunately, even though our service is very simple (two entity sets) we can't use this tool until the reference includes SAP throws in by default are all working and parseable.

[SAPODataV4Vocabularies.xml](https://github.com/OData/ODataConnectedService/files/5683961/SAPODataV4Vocabularies.xml.txt)

Tiberriver256:
Just an update.

I was able to get past the error about the 'BaseTerm' without modifying the csdl by selecting the option to ignore unknown attributes.

Also for better testing / repro here's the original root metadata.xml and all the reference models it includes:

[metadata.zip](https://github.com/OData/ODataConnectedService/files/5685132/metadata.zip)

PR #190, opened 2020-12-28
Excluding SAP common vocabulary namespaces

Tiberriver256:
Similar to the [Org.Odata namespace of vocabularies](https://github.com/oasis-tcs/odata-vocabularies/tree/master/vocabularies) the following should also all be excluded:
https://github.com/SAP/odata-vocabularies/tree/master/vocabularies

Fixes: #184

A better fix might be to add a new screen to the wizard which allows users to exclude specific namespaces from the codegen.


marabooy:
@gathogojr could you factor this effort in your current effort.
> A better fix might be to add a new screen to the wizard which allows users to exclude specific namespaces from the codegen.

g2mula:
I would prefer the _better fix_ approach, adding wizard page showing all referenced models with the core vocabulary ones already excluded...

Tiberriver256:
Sounds good. I'll ping you two again when I get it into this PR

PR #197, opened 2021-02-04
Fix warnings as a result of an empty metadata file generated when user opts to include T4 file

### LINKED ISSUE #195: Service proxy not generated and metadata file is empty when the option to include the T4 files is checked
### Describe the bug
When generating the service proxy and the user opts to include the T4 files, the service proxy file(s) are not generated and the metadata (csdl) file is empty.

ODataConnectedService v0.11.1
VisualStudio Version v16.8.4
Microsoft.Odata.Client v7.6.3

#### To Reproduce
1. Add a Connected Service to a project
2. On the **Settings** wizard page - **Advanced Settings** section, ensure that **Add code templates (Whether to include the T4 files into this project.)** checkbox is checked
3. Click Finish

#### Expected behavior
Service proxy file(s) to be generated and metadata file to be populated with the service metadata.

#### Actual behavior
Service proxy file(s) missing and the metadata file is empty.

#### Screenshots
image

gathogojr:
This pull request fixes #195

Fix warnings as a result of an empty metadata file generated when user opts to include T4 file.
By pre-populating the file with the root element to suppress the warnings. That content is later overwritten when the metadata is written to the file [here](https://github.com/OData/ODataConnectedService/blob/ef08797cbbf2e8b6809d88dd9142f95de61c8710/src/Templates/ODataT4CodeGenerator.cs#L4492) - `append: false` to overwrite the file

### DIFF
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -89,7 +89,7 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
 
             // When the T4 file is added to the target project, the proxy and metadata files 
             // are not automatically generated. To avoid ending up with an empty metadata file with 
-            // warnings, we pre-populate the it with the root element. The content will later be overwritten.
+            // warnings, we pre-populate it with the root element. The content will later be overwritten with the actual metadata when T4 template is run by the user.
             using (StreamWriter writer = File.CreateText(tempFile))
             {
                 await writer.WriteLineAsync("<edmx:Edmx Version=\"4.0\" xmlns:edmx=\"http://docs.oasis-open.org/odata/ns/edmx\">").ConfigureAwait(true);
--- a/src/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/CodeGeneration/V4CodeGenDescriptor.cs
@@ -63,13 +63,13 @@ namespace Microsoft.OData.ConnectedService.CodeGeneration
         {
             await Context.Logger.WriteMessageAsync(LoggerMessageCategory.Information, "Adding T4 files for OData V4...").ConfigureAwait(true);
 
-            var tempFile = Path.GetTempFileName();
+            var t4IncludeTempFile = Path.GetTempFileName();
             var t4Folder = Path.Combine(this.CurrentAssemblyPath, "Templates");
 
             var referenceFolder = this.GetReferenceFileFolder();
 
             // generate .ttinclude
-            using (StreamWriter writer = File.CreateText(tempFile))
+            using (StreamWriter writer = File.CreateText(t4IncludeTempFile))
             {
                 var ttIncludeText = File.ReadAllText(Path.Combine(t4Folder, "ODataT4CodeGenerator.ttinclude"));
                 if (this.TargetProjectLanguage == LanguageOption.GenerateVBCode)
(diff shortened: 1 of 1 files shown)


src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -87,18 +87,6 @@ private async Task AddT4FileAsync()
             var csdlFileName = string.Concat(ServiceConfiguration.ServiceName, Common.Constants.CsdlFileNameSuffix);
             var metadataFile = Path.Combine(referenceFolder, csdlFileName);
 
-            await Context.HandlerHelper.AddFileAsync(tempFile, metadataFile, new AddFileOptions() { SuppressOverwritePrompt = true }).ConfigureAwait(true);
```
KenitoInc, line 90: Why are we deleting this code?

gathogojr: @KenitoInc When the user opts to include the T4 file, the `TransformText()` method of `ODataT4CodeGenerator` class is NOT invoked. That method is responsible for generating service proxy code and saving it to disk as well as saving the service metadata to disk. Since its not invoked in the scenario described, we end up with an blank metadata file.
We had a discussion where we agreed to not to trigger the creation of metadata file - same way we don't create the service proxy file when the user checks that option. It will be created and populated when the user later runs the TT file from the client project.

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -87,18 +87,6 @@ private async Task AddT4FileAsync()
...
-
-            // Hack!
-            // Tests were failing since the test project cannot access ProjectItems
-            // dte == null when running test cases
-            var dte = VisualStudio.Shell.Package.GetGlobalService(typeof(DTE)) as DTE;
-            if (dte != null)
-            {
-                var projectItem = this.GetCsdlFileProjectItem(csdlFileName);
-                projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
-            }
```
habbes, line 100: When the T4 template generates the metadata file, does it create it as an embedded resource as well?

gathogojr: @habbes Investigated this and the answer is no. I'll fix that. Good catch. Thanks!

gathogojr: @habbes @KenitoInc Handled the issue differently by allowing the file to be generated but prepopulating with the root element to suppress the warnings

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -87,6 +87,15 @@ private async Task AddT4FileAsync()
             var csdlFileName = string.Concat(ServiceConfiguration.ServiceName, Common.Constants.CsdlFileNameSuffix);
             var metadataFile = Path.Combine(referenceFolder, csdlFileName);
 
+            // When the T4 file is added to the target project, the proxy and metadata files 
+            // are not automatically generated. To avoid ending up with an empty metadata file with 
+            // warnings, we pre-populate the it with the root element. The content will later be overwritten.
```
habbes, line 92:
```suggestion
            // warnings, we pre-populate it with the root element. The content will later be overwritten with the actual metadata when T4 template is run by the user.
```

src/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -99,6 +108,8 @@ private async Task AddT4FileAsync()
                 projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
             }
 
+            tempFile = Path.GetTempFileName();
+
```
habbes, line 112: Why are we regenerating the `tempFile` here?

habbes: Oh I see, we're re-using the variable `tempFile` for different files. Maybe we should have had different variable names to avoid confusion.

habbes: It's a bid hard to follow now when `tempFile` refers to different files and hard to tell from the immediate context. Maybe we should have had `tempTtIncludeFile`, `tempMetadataFile` and `tempTtFile`. Or `sourceTtIncludeFile`, etc. since the temp files are the source files passed to the `AddFileAsync` method.

habbes (approved):
Left some minor refactoring suggestions. But other than that, looks good to me.

marabooy (approved):

PR #202, opened 2021-02-18
Fix #201 : Escape quotes not applied in Private Const Edmx As String

### LINKED ISSUE #201: Escape quotes not applied in Private Const Edmx As String
When using an OData api endpoint the client code genrator is VB.NET forgets to set the escape quotes in the Edmx XML string.

For example, this is generated:
```xml
Private Const Edmx As String = "<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">" &
```
Which doesn't work offcourse. This is what i expect (and does work)
```xml
Private Const Edmx As String = "<edmx:Edmx Version=""4.0"" xmlns:edmx=""http://docs.oasis-open.org/odata/ns/edmx"">" &
```

unchase:
Fix #201.

Added the tests below:

To **ODataT4CodeGeneratorTests:**

- [x] `CodeGenUsingTempMetadataFileForVBTest`.


unchase:

gathogojr:
@unchase Does your PR fix the wrong escaping of the string const in VB or something different?

unchase:
> @unchase Does your PR fix the wrong escaping of the string const in VB or something different?

Yes, PR fix the wrong escaping of the string const in VB

gathogojr:
> > @unchase Does your PR fix the wrong escaping of the string const in VB or something different?
> 
> Yes, PR fix the wrong escaping of the string const in VB

I was struggling to understand how it does that but I figured it out after another look

gathogojr (approved):

marabooy (approved):

KenitoInc (approved):
LGTM

PR #207, opened 2021-03-03
Fix #206 : VB.Net generated code does not use NameCsdl.xml

### LINKED ISSUE #206: VB.Net generated code does not use NameCsdl.xml
When using the generator for VB.NET the client EDM XML is added in the root class as an XML string under variable `Private Const Edmx As String = ` but in C# this generates an XML file called `MyNameCsdl.xml` When a new instance is created the XML file is read. But in VB.NET the String XML is used and the file is created but contains nothing. Can this work the same for VB.NET as is used for C#?

unchase:
Fix #206

### DIFF
--- a/src/Templates/ODataT4CodeGenerator.cs
+++ b/src/Templates/ODataT4CodeGenerator.cs
@@ -2206,13 +2206,20 @@ public abstract class ODataClientTemplate : TemplateBase
             }
         }
 
-        if (useTempFile && this.context.TargetLanguage == LanguageOption.CSharp)
+        if (useTempFile)
         {
             this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx));
         }
         else
         {
-            this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\""));
+            if (this.context.TargetLanguage == LanguageOption.VB)
+            {
+                this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\"").Replace("\r\n", "\" & _\r\n \""));
+            }
+            else
+            {
+                this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\""));
+            }
         }
 
         bool hasOperationImport = container.OperationImports().OfType<IEdmOperationImport>().Any();
@@ -6562,7 +6569,17 @@ this.Write(")\r\n        End Sub\r\n");
 
     internal override void WriteGeneratedEdmModel(string escapedEdmxString)
     {
-        escapedEdmxString = escapedEdmxString.Replace("\r\n", "\" & _\r\n \"");
+        string path = this.context.MetadataFilePath;
+        string relativePath = this.context.MetadataFileRelativePath;
+        if(!String.IsNullOrEmpty(path))
+        {
+            using (StreamWriter writer = new StreamWriter(path, false))
+            {
+                writer.WriteLine(escapedEdmxString);
+            }
+        }
+
+        bool useTempFile = !String.IsNullOrEmpty(path) && System.IO.File.Exists(path);
 
 this.Write("        <Global.System.CodeDom.Compiler.GeneratedCodeAttribute(\"Microsoft.OData.C" +
         "lient.Design.T4\", \"");
@@ -6617,8 +6634,30 @@ this.Write("            <Global.System.CodeDom.Compiler.GeneratedCodeAttribute(\
 this.Write(this.ToStringHelper.ToStringWithCulture(T4Version));
 
 this.Write("\")>  _\r\n            Private Shared ParsedModel As Global.Microsoft.OData.Edm.IEdm" +
-        "Model = LoadModelFromString\r\n            <Global.System.CodeDom.Compiler.Generat" +
-        "edCodeAttribute(\"Microsoft.OData.Client.Design.T4\", \"");
+        "Model = LoadModelFromString\r\n");
+
+
+            if (useTempFile)
+            {
+
+this.Write("            <Global.System.CodeDom.Compiler.GeneratedCodeAttribute(\"Microsoft.ODa" +
+        "ta.Client.Design.T4\", \"");
+
+this.Write(this.ToStringHelper.ToStringWithCulture(T4Version));
+
+this.Write("\")>  _\r\n            Private Const filePath As String = \"");
+
+this.Write(this.ToStringHelper.ToStringWithCulture(relativePath));
+
+this.Write("\"\r\n");
+
+
+            }
+            else
+            {
+
+this.Write("            <Global.System.CodeDom.Compiler.GeneratedCodeAttribute(\"Microsoft.ODa" +
+        "ta.Client.Design.T4\", \"");
 
 this.Write(this.ToStringHelper.ToStringWithCulture(T4Version));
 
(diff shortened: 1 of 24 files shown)


unchase:

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2188,13 +2188,20 @@ internal void WriteEntityContainer(IEdmEntityContainer container, string fullNam
             }
         }
 
-        if (useTempFile && this.context.TargetLanguage == LanguageOption.CSharp)
```
gathogojr, line 2191: @unchase Did you not make this change in this [other PR](https://github.com/OData/ODataConnectedService/pull/202/files#diff-d022877b2d519bc2c2c5e013aaeb5242ee113efc88980eb064299ef6a065ea10R2191) to ensure this block is not executed when the language is VB? Won't this change undo the previous one?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2188,13 +2188,20 @@ internal void WriteEntityContainer(IEdmEntityContainer container, string fullNam
...
+        if (useTempFile)
         {
             this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx));
         }
         else
         {
-            this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\""));
+            if (this.context.TargetLanguage == LanguageOption.VB)
+            {
+                this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\"").Replace("\r\n", "\" & _\r\n \""));
```
gathogojr, line 2199: @unchase What does this expression represent: `"\" & _\r\n \""`?
As a test,
```csharp
Console.WriteLine("Test\r\n".Replace("\r\n", "\" & _\r\n \""))
```
yields the following:
```
Test" & _
 "
```
Is that the behaviour or kind of result we're looking for, or does the replacement make sense when considered together with `.Replace("\"", "\"\"")`?

src/Templates/ODataT4CodeGenerator.cs
gathogojr: @unchase Or maybe this change [here](https://github.com/OData/ODataConnectedService/pull/207/files#diff-d022877b2d519bc2c2c5e013aaeb5242ee113efc88980eb064299ef6a065ea10L6547) takes care of that concern...

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2188,13 +2188,20 @@ internal void WriteEntityContainer(IEdmEntityContainer container, string fullNam
...
 
-        if (useTempFile && this.context.TargetLanguage == LanguageOption.CSharp)
+        if (useTempFile)
         {
             this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx));
         }
         else
         {
-            this.WriteGeneratedEdmModel(Utils.SerializeToString(this.context.Edmx).Replace("\"", "\"\""));
+            if (this.context.TargetLanguage == LanguageOption.VB)
```
gathogojr, line 2197: @unchase We could move the call to `Utils.SerializeToString(this.context.Edmx)` outside the `if` and `else` blocks such that we don't have to repeat it 3 times - place the result in a variable and make use of that instead.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2188,13 +2188,20 @@ internal void WriteEntityContainer(IEdmEntityContainer container, string fullNam
             }
         }
 
-        if (useTempFile && this.context.TargetLanguage == LanguageOption.CSharp)
+        if (useTempFile)
```
gathogojr, line 2191: @unchase A change may not be necessary for now and it needn't be in this PR but someone might argue that since `useTempFile` is defined [here](https://github.com/OData/ODataConnectedService/pull/207/files#diff-d022877b2d519bc2c2c5e013aaeb5242ee113efc88980eb064299ef6a065ea10R2098) and used only on this line, and then in `WriteGeneratedModel` method we still check if temp file should be used [here](https://github.com/OData/ODataConnectedService/pull/207/files#diff-d022877b2d519bc2c2c5e013aaeb5242ee113efc88980eb064299ef6a065ea10R6556) and [here](https://github.com/OData/ODataConnectedService/pull/207/files#diff-d022877b2d519bc2c2c5e013aaeb5242ee113efc88980eb064299ef6a065ea10R6564) and a couple other places, then maybe the variable belongs in that method.

test/ODataConnectedService.Tests/CodeGenReferences/AbstractEntityTypeWithoutKey.vb
```diff
@@ -12,17 +12,17 @@ Option Strict Off
...
 
-'Generation date: 18.03.2020 15:06:23
+'Generation date: 03.03.2021 1:44:57
 Namespace AbstractEntityTypeWithoutKey
-    '''<summary>
-    '''There are no comments for Container in the schema.
-    '''</summary>
+        ''' <summary>
+        ''' There are no comments for Container in the schema.
+        ''' </summary>
```
gathogojr, line 19:
```suggestion
    ''' </summary>
```
I think the indentation on the docstrings preceding class definitions are off.

test/ODataConnectedService.Tests/CodeGenReferences/EntitiesEnumsFunctionsSelectTypes.vb
```diff
@@ -12,17 +12,17 @@ Option Strict Off
...
     Partial Public Class Container
         Inherits Global.Microsoft.OData.Client.DataServiceContext
-        '''<summary>
-        '''Initialize a new Container object.
-        '''</summary>
-        <Global.System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "#VersionNumber#")>  _
+        ''' <summary>
+        ''' Initialize a new Container object.
+        ''' </summary>
+        <Global.System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "2.4.0")>  _
```
gathogojr, line 25: @unchase Hopefully pinning the version to 2.4.0 would never cause problems to some users. There are files where its pinned to 2.4.0 and it wasn't your change so it should be okay I think.

src/Templates/ODataT4CodeGenerator.cs
unchase: Second)

src/Templates/ODataT4CodeGenerator.cs
unchase: It's need to be used for correct constructing `Private Const Edmx As String =` in generated VB files.

For example (for generated `OverrideOperations.vb` file):

image

Yellow - `.Replace("\"", "\"\"")`
green - `.Replace("\r\n", "\" & _\r\n \"")`

src/Templates/ODataT4CodeGenerator.cs
unchase: Agree with you

src/Templates/ODataT4CodeGenerator.cs
unchase: I do not quite understand what exactly you are proposing)

test/ODataConnectedService.Tests/CodeGenReferences/EntitiesEnumsFunctionsSelectTypes.vb
unchase: It doesn't matter which version is used in the generated files. All tests ignore this value.

test/ODataConnectedService.Tests/CodeGenReferences/AbstractEntityTypeWithoutKey.vb
unchase: The only one method `WriteDescriptionSummary` with **fixed indentation** used in the generation process for all summary descriptions.
Therefore some of the indentation are wrong.

unchase: Fixed in https://github.com/OData/ODataConnectedService/pull/207/commits/7a68d2f3b5330ed0948bbaba2a7b23bc3df0b1ea

gathogojr (approved):
LGTM

marabooy (approved):

PR #214, opened 2021-03-08
Fix #205 : DTO attributes are ignored

### LINKED ISSUE #205: DTO attributes are ignored
### Describe the bug
`Edm.string` with` MaxLength` or `Nullable="false"` do not affect generated DTO.

ODataConnectedService v0.12.x
VisualStudio Version v16.8.5
Microsoft.Odata.Client v7.8.2

#### To Reproduce
Create reference to OData service with an entity with a property of type `Edm.string` and a `MaxLength` or `Nullable="false"` property.

#### Expected behavior
Generated DTO should have a `MaxLength` = to the `MaxLength` in OData metadata and `Required`= to `!Nullable`.

#### Actual behavior
Generated DTO is the same, no matter what values `MaxLength `and `Nullable` have.

#### Additional context
Example metadata:
```
      <EntityType Name="dimension">
        <Key>
          <PropertyRef Name="dimensionCode" />
          <PropertyRef Name="code" />
        </Key>
        <Property Name="dimensionCode" Type="Edm.String" Nullable="false" MaxLength="20" />
        <Property Name="code" Type="Edm.String" Nullable="false" MaxLength="20" />
        <Property Name="name" Type="Edm.String" MaxLength="50" />
      </EntityType>
```

Generated DTO excerpt:

```
        /// <summary>
        /// There are no comments for Property Name in the schema.
        /// </summary>
        [global::Microsoft.OData.Client.OriginalNameAttribute("name")]
        public virtual string Name
        {
            get
            {
                return this._Name;
            }
            set
            {
                this.OnNameChanging(value);
                this._Name = value;
                this.OnNameChanged();
                this.OnPropertyChanged("name");
            }
        }
        private string _Name;
        partial void OnNameChanging(string value);
        partial void OnNameChanged();
        /// <summary>
        /// This event is raised when the value of the property is changed
        /// </summary>

```

unchase:
Fix #205


marabooy:
@unchase I have merged #207

unchase:
> @unchase I have merged #207

Thanks, I will try to finish the PR this week.

unchase:

marabooy (approved):

unchase:
@habbes @marabooy @KenitoInc @ElizabethOkerio @gathogojr @paulodero rebased

marabooy:
Thanks @unchase taking a look at this.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1666,7 +1666,7 @@ internal string SingleSuffix
     internal abstract void WriteParameterNullCheckForStaticCreateMethod(string parameterName);
     internal abstract void WritePropertyValueAssignmentForStaticCreateMethod(string instanceName, string propertyName, string parameterName);
     internal abstract void WriteMethodEndForStaticCreateMethod(string instanceName);
-    internal abstract void WritePropertyForStructuredType(string propertyType, string originalPropertyName, string propertyName, string fixedPropertyName, string privatePropertyName, string propertyInitializationValue, string propertyAttribute, string propertyDescription, bool writeOnPropertyChanged, IDictionary<string, string> revisionAnnotations);
+    internal abstract void WritePropertyForStructuredType(string propertyType, string originalPropertyName, string propertyName, string fixedPropertyName, string privatePropertyName, string propertyInitializationValue, string propertyAttribute, string propertyDescription, int? propertyMaxLength, bool writeOnPropertyChanged, bool isNullable, IDictionary<string, string> revisionAnnotations);
```
habbes, line 1669: I don't think adding more and more parameters to this method for every new attributes we want to support is a sustainable approach. In my view we should have a class where all these annotations could be stored as properties, and then we just pass a single object that encapsulate all attributes. That way whenever we want to support a new attribute, we just add a property to the class and only update the methods where that attribute is relevant, instead of adding new arguments to so many places.

habbes (approved):

habbes:
@unchase do you have any test case for the `StringLengthAttribute`? I've seen some for the `RequiredAttribute`

src/Templates/ODataT4CodeGenerator.cs
unchase: Agree with you. I also thought about it.

unchase:
> @unchase do you have any test case for the `StringLengthAttribute`? I've seen some for the `RequiredAttribute`

I'll add them tonight.

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
```diff
@@ -591,9 +591,9 @@ internal override void WriteMethodEndForStaticCreateMethod(string instanceName)
...
+            internal override void WritePropertyForStructuredType(string propertyType, string originalPropertyName, string propertyName, string fixedPropertyName, string privatePropertyName, string propertyInitializationValue, string propertyAttribute, string propertyDescription, int? propertyMaxLength, bool writeOnPropertyChanged, bool isNullable, IDictionary<string, string> revisionAnnotations)
             {
-                this.CalledActions.Add("WritePropertyForStructuredType(" + propertyType + ", " + originalPropertyName + ", " + propertyName + ", " + fixedPropertyName + ", " + privatePropertyName + ", " + propertyInitializationValue + ", " + propertyAttribute + ", " + writeOnPropertyChanged + ", " + revisionAnnotations + ")");
+                this.CalledActions.Add("WritePropertyForStructuredType(" + propertyType + ", " + originalPropertyName + ", " + propertyName + ", " + fixedPropertyName + ", " + privatePropertyName + ", " + propertyInitializationValue + ", " + propertyAttribute + ", " + propertyMaxLength + ", " + writeOnPropertyChanged + ", " + isNullable + ", " + revisionAnnotations + ")");
```
wandeg, line 596: Any reason not to use [String.Join](https://docs.microsoft.com/en-us/dotnet/api/system.string.join?view=net-5.0#System_String_Join_System_String_System_String___) ?

unchase: We can use `PropertyOptions` class instead of it.

unchase: Added

src/Templates/ODataT4CodeGenerator.cs
unchase: Fixed

wandeg (approved):

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5218,7 +5296,7 @@ internal override void WriteEnumFlags()
 
     internal override void WriteSummaryCommentForEnumType(string enumName, string description)
     {
-        WriteDescriptionSummary(string.IsNullOrWhiteSpace(description) ? $"There are no comments for {enumName} in the schema." : description);
+        WriteDescriptionSummary(string.IsNullOrWhiteSpace(description) ? $"There are no comments for {enumName} in the schema." : description, true);
```
KenitoInc, line 5299: Why did we make `isClass = true`?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -7333,7 +7450,7 @@ internal override void WriteEnumFlags()
...
```
KenitoInc, line 7453: Same comment as above regarding `isClass = true`.

KenitoInc (approved):

KenitoInc:
Thanks for this PR @unchase

PR #217, opened 2021-03-29
Add Obsolete Attributes

KenitoInc:
Add support for Obsolete Attributes in the generate code.

Here is how the annotation looks in XML. It can be applied to a type, entity set, singleton, property, navigation property, function or action. If a type is marked as deprecated, it is not necessary to mark members of that type as deprecated, nor is it necessary to annotate any usage of that type in entity sets, singletons, properties, navigation properties, functions, or actions.

```xml
  <EntityType Name="outlookTask" BaseType="Microsoft.OutlookServices.outlookItem" ags:IsMaster="true" ags:WorkloadName="Task" ags:EnabledForPassthrough="true">
    <Annotation Term="Org.OData.Core.V1.Revisions">
      <Collection>
        <Record>
          <PropertyValue Property = "Date" Date="2020-08-20"/>
          <PropertyValue Property = "Version" String="2020-08/Tasks_And_Plans"/>
          <PropertyValue Property = "Kind" EnumMember="Org.OData.Core.V1.RevisionKind/Deprecated"/>
          <PropertyValue Property = "Description" String="The Outlook tasks API is deprecated and will stop returning data on August 20, 2022. Please use the new To Do API."/>
          <PropertyValue Property = "RemovalDate" Date="2022-08-20"/>
        </Record>
      </Collection>
    </Annotation>
    ...
  </EntityType>
```xml

### CODE REVIEW

File: src/Templates/ODataT4CodeGenerator.cs [hunk 1]
```diff
@@ -3077,6 +3083,29 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+
+    /// <summary>
+    /// Searches through model's vocabulary annotations and returns annotation of term 'Revisions' if present
+    /// </summary>
+    private IEdmStringConstantExpression GetRevisionsAnnotation(IEdmVocabularyAnnotatable model)
+    {
+        IEnumerable<IEdmExpression> collection = model.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.Name == "Revisions")
+            .Select(x => x.Value);
+
+        if (collection != null && collection.Count() > 0)
```
habbes, line 3095:
```suggestion
        if (collection?.Any())
```

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3083,29 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
         return model.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.Name == "Description")
             .Select(x => x.Value).FirstOrDefault() as IEdmStringConstantExpression;
     }
+
+    /// <summary>
+    /// Searches through model's vocabulary annotations and returns annotation of term 'Revisions' if present
+    /// </summary>
+    private IEdmStringConstantExpression GetRevisionsAnnotation(IEdmVocabularyAnnotatable model)
+    {
+        IEnumerable<IEdmExpression> collection = model.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.Name == "Revisions")
```
habbes, line 3092: Why aren't you also checking the namespace of the annotation term?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3083,29 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+            .Select(x => x.Value);
+
+        if (collection != null && collection.Count() > 0)
+        {
+            IEdmCollectionExpression semanticElement = collection.FirstOrDefault() as IEdmCollectionExpression;
+            IEdmRecordExpression element = semanticElement.Elements.FirstOrDefault() as IEdmRecordExpression;
```
habbes, line 3098:
```suggestion
            IEdmRecordExpression element = semanticElement?.Elements.FirstOrDefault() as IEdmRecordExpression;
```

Since you're using `FirstOrDefault()`, I think you should check for nulls, just in case.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3083,29 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+            IEdmStringConstantExpression description = element.Properties.Where(x => x.Name == "Description").Select(x => x.Value).FirstOrDefault() as IEdmStringConstantExpression;
```
habbes, line 3099:
```suggestion
            IEdmStringConstantExpression description = element?.Properties.Where(x => x.Name == "Description").Select(x => x.Value).FirstOrDefault() as IEdmStringConstantExpression;
```

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3083,29 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
         return model.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.Name == "Description")
             .Select(x => x.Value).FirstOrDefault() as IEdmStringConstantExpression;
     }
+
+    /// <summary>
+    /// Searches through model's vocabulary annotations and returns annotation of term 'Revisions' if present
+    /// </summary>
+    private IEdmStringConstantExpression GetRevisionsAnnotation(IEdmVocabularyAnnotatable model)
```
habbes, line 3090: Maybe rename `model` parameter to something else, like `schemaElement` to avoid confusion with Edm model.

habbes: Out of curiosity, why are we returning the `StringConsantExpression` instead of just returning the description as a `string`?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1645,8 +1645,8 @@ internal string SingleSuffix
     internal abstract void WriteMethodStartForResolveNameFromType(string containerName, string fullNamespace);
     internal abstract void WriteResolveType(string fullNamespace, string languageDependentNamespace);
     internal abstract void WriteMethodEndForResolveNameFromType(bool modelHasInheritance);
-    internal abstract void WriteContextEntitySetProperty(string entitySetName, string entitySetFixedName, string originalEntitySetName, string entitySetElementTypeName, string description, bool inContext = true);
-    internal abstract void WriteContextSingletonProperty(string singletonName, string singletonFixedName, string originalSingletonName, string singletonElementTypeName, string description, bool inContext = true);
+    internal abstract void WriteContextEntitySetProperty(string entitySetName, string entitySetFixedName, string originalEntitySetName, string entitySetElementTypeName, string description, string revisionDescription, bool inContext = true);
```
habbes, line 1648: I think we have too many parameters on some of these Write methods. I don't think it's sustainable or maintainable to keep adding more.. I suggest we use a dictionary are an object to hold all the additional metadata like descriptions, deprecation messages, etc.:

```c#
class CodeElementMetadata {
   public string Description { get; set; }
   public string DeprecationMessage { get; set; }
}
```
Not sure about the name of the class, or whether we should have a common class for all types of code elements (entity sets, methods, etc.) or different classes.

habbes (changes requested):

src/Templates/ODataT4CodeGenerator.cs
ElizabethOkerio, line 3092: Add a check for the Revision Kind as well because we can have other Revision Kinds (Added, Modified) in the model

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5005,16 +5036,17 @@ internal override void WriteMethodEndForStaticCreateMethod(string instanceName)
...
         WriteDescriptionSummary(string.IsNullOrWhiteSpace(propertyDescription) ? $"There are no comments for Property {propertyName} in the schema." : propertyDescription);
+        WriteRevisionDescription(revisionDescription);
 
 this.Write("        [global::System.CodeDom.Compiler.GeneratedCodeAttribute(\"Microsoft.OData." +
         "Client.Design.T4\", \"");
 
 this.Write(this.ToStringHelper.ToStringWithCulture(T4Version));
 
-this.Write("\")]\r\n");
+this.Write("\")]\r\n\r\n");
```
marabooy, line 5049:
```suggestion
this.Write("\")]\r\n\");
```

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5915,6 +5953,37 @@ protected override void WriteDescriptionSummary(string description, bool isClass
...
+    {
+        if (string.IsNullOrEmpty(revisionDescription))
+        {
+            return;
+        }
+
+        if (isClass)
+        {
+
+this.Write("    [global::System.ObsoleteAttribute(\"");
```
marabooy, line 5966: We could probably just write this using one invocation to this.Write and string.Concat

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5915,6 +5953,37 @@ protected override void WriteDescriptionSummary(string description, bool isClass
...
+this.Write(this.ToStringHelper.ToStringWithCulture(revisionDescription));
+
+this.Write("\")]\r\n");
+
+
+        }
+        else
+        {
+
+this.Write("        [global::System.ObsoleteAttribute(\"");
```
marabooy, line 5977: Same for string.Concat to reduce LOC

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I have fix by using `collection != null && collection.Any()`

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I changed to use a `Dictionary` that stores all revision annotations.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I don't have control on how the generated code is written.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I don't have control on how the generated code is written.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I don't have control on how the generated code is written.

src/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2804,6 +2810,9 @@ public abstract class ODataClientTemplate : TemplateBase
                 containerPropertyAttribute = this.ContainerPropertyAttribute;
             }
 
+            // Add empty dictionary to the anonymous type
+            IDictionary<string, string> emptyDict = new Dictionary<string, string>();
+
```
habbes, line 2815: Could you explain what this is for?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2950,6 +2956,9 @@ internal void WritePropertiesForStructuredType(IEdmStructuredType structuredType
                 containerPropertyAttribute = this.ContainerPropertyAttribute;
             }
 
+            // Add empty dictionary to the anonymous type
```
habbes, line 2959: Which anonymous type are you referring to here?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2959,7 +2968,8 @@ internal void WritePropertiesForStructuredType(IEdmStructuredType structuredType
                 PrivatePropertyName = "_" + containerPropertyName,
                 PropertyInitializationValue = string.Format(this.DictionaryConstructor, this.StringTypeName, this.ObjectTypeName),
                 PropertyAttribute = containerPropertyAttribute,
-                PropertyDescription = string.Empty
+                PropertyDescription = string.Empty,
+                RevisionDescription = emptyDict
```
habbes, line 2972: Why not create the empty dict here: `RevisionDescription = new Dictionary<string, string>()`?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2980,7 +2990,8 @@ internal void WritePropertiesForStructuredType(IEdmStructuredType structuredType
                 propertyInfo.PropertyInitializationValue,
                 propertyInfo.PropertyAttribute,
                 propertyInfo.PropertyDescription,
-                useDataServiceCollection);
+                useDataServiceCollection,
+                propertyInfo.RevisionDescription);
```
habbes, line 2994: Logically, I think it would make more sense to group the property metadata-related parameters together, then the `useDataServiceCollection` last. But I still think adding more params is unsustainable.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3088,33 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+    /// Searches through model's vocabulary annotations and returns dictionary of term 'Revisions' annotation kinds.
+    /// </summary>
+    private IDictionary<string,string> GetRevisionsAnnotation(IEdmVocabularyAnnotatable schemaElement)
+    {
+        IEnumerable<IEdmExpression> collection = schemaElement.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.FullName() == "Org.OData.Core.V1.Revisions")
+            .Select(x => x.Value);
+
+        IDictionary<string, string> revisionsAnnotation = new Dictionary<string, string>();
+
+        if (collection != null && collection.Any())
```
habbes, line 3102:
```suggestion
        if (collection?.Any())
```

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3088,33 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+        {
+            IEdmCollectionExpression semanticElement = collection.FirstOrDefault() as IEdmCollectionExpression;
```
habbes, line 3104: Maybe you can verify that `semanticElement` is not null since you're using `FirstOrDefault()`?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5915,6 +5962,37 @@ protected override void WriteDescriptionSummary(string description, bool isClass
         }
     }
 
+    protected override void WriteObsoleteAttribute(IDictionary<string, string> revisionAnnotations, bool isClass = false)
+    {
+        if (!revisionAnnotations.TryGetValue("Deprecated", out string revisionDescription))
```
habbes, line 5967: Maybe "Deprecated" should be a constant?

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3088,33 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+        if (collection != null && collection.Any())
+        {
+            IEdmCollectionExpression semanticElement = collection.FirstOrDefault() as IEdmCollectionExpression;
+
+            foreach (IEdmRecordExpression element in semanticElement.Elements)
+            {
+                string description = (element?.Properties.Where(x => x.Name == "Description").Select(x => x.Value).FirstOrDefault() as IEdmStringConstantExpression).Value;
+                IEdmEnumMemberExpression revisionKind = element?.Properties.Where(x => x.Name == "Kind").Select(x => x.Value).FirstOrDefault() as IEdmEnumMemberExpression;
+                string name = revisionKind?.EnumMembers.FirstOrDefault().Name;
+                revisionsAnnotation.Add(name, description);
```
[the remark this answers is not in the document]
habbes: Is there a guarantee the `FirstOrDefault()` cannot return `null`? What if the revision kind is missing from the xml document?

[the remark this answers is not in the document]
odero: Hence the second part of my comment. Kind is a required field. Meaning if it returns `null` it should be an error which should be handled/propagated appropriately and/or reported

src/Templates/ODataT4CodeGenerator.ttinclude
KenitoInc: Since we are not adding `ObsoleteAttribute` to dynamic types, we are assigning an empty dictionary to the property in the anonymous type.
Note: An anonymous type cannot have a null property.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: `propertyInfo`
Check the entire `internal void WritePropertiesForStructuredType(IEdmStructuredType structuredType)` method.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: You can't do that in anonymous types.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: I have updated the property name,
All `RevisionAnnotations` are in one dictionary.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: `collection?.Any()` returns a `bool?`. Unless I add an explicit cast as below to get a `bool`
`(bool)(collection?.Any())`

odero: if (collection?.Any() == true)

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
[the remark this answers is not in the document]
KenitoInc: I don't there is a config for failing silently.
If required fields are missing, we should throw an exception.
I have updated the code appropriately and added tests.

test/ODataConnectedService.Tests/CodeGenReferences/EntityAndNavPropertiesMarkedObsolete.xml
```diff
@@ -0,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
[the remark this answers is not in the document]
KenitoInc: Done

habbes (approved):

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2147,7 +2150,9 @@ internal void WriteEntityContainer(IEdmEntityContainer container, string fullNam
                 camelCaseEntitySetName = Customization.CustomizeNaming(camelCaseEntitySetName);
             }
 
-            this.WriteContextEntitySetProperty(camelCaseEntitySetName, GetFixedName(camelCaseEntitySetName), entitySet.Name, GetFixedName(entitySetElementTypeName), GetDescriptionAnnotation(entitySet)?.Value);
+            IDictionary<string, string> revisionAnnotations = new Dictionary<string, string>();
+
+            this.WriteContextEntitySetProperty(camelCaseEntitySetName, GetFixedName(camelCaseEntitySetName), entitySet.Name, GetFixedName(entitySetElementTypeName), GetDescriptionAnnotation(entitySet)?.Value, GetRevisionsAnnotation(entitySet));
```
gathogojr, line 2155: @KenitoInc The naming is a bit confusing. You changed the method signature to accept an additional parameter `revisionAnnotations` then adopted `GetRevisionsAnnotation` as the name for method that returns that.

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2935,7 +2942,8 @@ internal void WritePropertiesForStructuredType(IEdmStructuredType structuredType
                     PrivatePropertyName = "_" + propertyName,
                     PropertyInitializationValue = Utils.GetPropertyInitializationValue(property, useDataServiceCollection, this, this.context),
                     PropertyAttribute = string.Empty,
-                    PropertyDescription = GetDescriptionAnnotation(property)?.Value
+                    PropertyDescription = GetDescriptionAnnotation(property)?.Value,
+                    RevisionAnnotations = GetRevisionsAnnotation(property)
```
gathogojr, line 2946: Related to my other comment https://github.com/OData/ODataConnectedService/pull/217/files#r633241743
LHS you have annotations while RHS you have annotation
```csharp
RevisionAnnotations = GetRevisionsAnnotation(property)
```

src/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3077,6 +3090,48 @@ private IEdmStringConstantExpression GetDescriptionAnnotation(IEdmVocabularyAnno
...
+    private IDictionary<string,string> GetRevisionsAnnotation(IEdmVocabularyAnnotatable schemaElement)
+    {
+        IEnumerable<IEdmExpression> collection = schemaElement.VocabularyAnnotations(this.context.EdmModel).Where(x => x.Term.FullName() == "Org.OData.Core.V1.Revisions")
+            .Select(x => x.Value);
+
+        IDictionary<string, string> revisionsAnnotation = new Dictionary<string, string>();
+
+        if (collection?.Any() == true)
+        {
+            IEdmCollectionExpression semanticElement = collection.FirstOrDefault() as IEdmCollectionExpression;
```
gathogojr, line 3106:
```suggestion
        IEdmCollectionExpression semanticElement = collection?.FirstOrDefault() as IEdmCollectionExpression;
```
Since `.Any()` is always executed whenever the collection is not null, I think you can forgo that and just call `.FirstOrDefault()`

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTests.cs
```diff
@@ -748,7 +822,9 @@ private static void GeneratedCodeShouldCompile(string source, bool isCSharp)
             }
 
             var results = codeProvider.CompileAssemblyFromSource(compilerOptions, source);
-            results.Errors.Should().BeEmpty();
+
+            // TODO: Activate when we fix why warnings are still being treated as errors.
+            // results.Errors.Should().BeEmpty();
```
gathogojr, line 827: Since you changed the `TreatWarningsAsErrors` setting to `false`, should you still leave this line commented?

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTests.cs
```diff
@@ -579,6 +579,80 @@ public void CodeGenSelectingSchemaTypesTest()
...
+        {
+            string code = CodeGenWithT4Template(ODataT4CodeGeneratorTestDescriptors.SingletonsMarkedObsolete.Metadata, null, true, false);
+            ODataT4CodeGeneratorTestDescriptors.SingletonsMarkedObsolete.Verify(code, true/*isCSharp*/, false/*useDSC*/);
+
+            code = CodeGenWithT4Template(ODataT4CodeGeneratorTestDescriptors.SingletonsMarkedObsolete.Metadata, null, false, false);
+            ODataT4CodeGeneratorTestDescriptors.SingletonsMarkedObsolete.Verify(code, false/*isCSharp*/, false/*useDSC*/);
+        }
+
+        [TestMethod]
+        public void RevisionsAnnotationMissingRevisionKindThrows()
```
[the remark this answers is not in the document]
KenitoInc: It's not a breaking since this will be only thrown when you have an incorrect AnnotationXml.
`Kind` and `Description` are required properties.

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

src/Templates/ODataT4CodeGenerator.cs
KenitoInc: #Resolved

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTests.cs
KenitoInc: I have uncommented the line.
However still some warnings are showing up as errors. I have opted to set the warning level as 0 (to ignore all warnings) and check for compilation errors only.

gathogojr (approved):
:shipit:

PR #236, opened 2021-11-29
generate proxy command options

ElizabethOkerio:
This PR contains a list of options the <code> odata-cli </code> generate command will have.

### DIFF
--- a/src/OData.Cli/OData.Cli.csproj
+++ b/src/OData.Cli/OData.Cli.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.1.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21308.1" />
   </ItemGroup>
 
 </Project>
--- a/src/OData.Cli/Program.cs
+++ b/src/OData.Cli/Program.cs
@@ -1,18 +1,88 @@
-﻿using Microsoft.Extensions.CommandLineUtils;
+﻿using System.CommandLine;
+using System.Threading.Tasks;
 
 namespace OData.Cli
 {
     class Program
     {
-        static void Main(string[] args)
+        static async Task Main(string[] args)
         {
-            var app = new CommandLineApplication();
+            var app = new RootCommand
+            {
+                new Command("generate", "Command to generate proxy classes for OData endpoints.")
+                {
+                    new Option<string>(
+                        new[] { "--metadata|-m"},
+                        "The URI of the metadata document. The value must be set to a valid service document URI or a local file path."
+                    ){IsRequired = true},
+                    new Option<string>(
+                        new[] { "--load-settings-from-file|-lsff" },
+                        "The location of the configuration settings file."
+                    ),
+                    new Option<string>(
+                        new[] { "--custom-headers|-h"},
+                        "Headers that will get sent along with the request when fetching the metadata document from the service. Format: Header1:HeaderValue, Header2:HeaderValue"
+                    ),
+                    new Option<string>(
+                        new[] { "--proxy|-p"},
+                        "Proxy settings. Format: domain\\user:password@SERVER:PORT."
+                    ),
+                    new Option<string>(
+                        new[] { "--namespace|-ns"},
+                        "The namespace of the client code generated. Example:ODataCliCodeGeneratorSample.NorthWindModel or ODataCliCodeGeneratorSample or it could be a name related to the OData endpoint."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-tracking|-n"},
+                        "Disables entity and property tracking. Entity tracking enables the client to track an entity in DataServiceContext while property tracking enables the client to track only top-level properties."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--upper-camel-case|-ucc"},
+                        "Disables upper camel casing."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--internal|-i"},
+                        "Apply the \"internal\" class modifier on generated classes instead of \"public\" thereby making them invisible outside the assembly."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--multiple-files|-mf"},
+                        "Split the generated classes into separate files instead of generating all the code in a single file."
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-operation-imports|-eoi"},
+                        "Comma-separated list of the names of operation imports to exclude from the generated code. Example: ExcludedOperationImport1,ExcludedOperationImport2"
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-bound-operations|-ebo"},
+                        "Comma-separated list of the names of bound operations to exclude from the generated code.Example: BoundOperation1,BoundOperation2"
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-entity-types|-est"},
+                        "Comma-separated list of the names of entity types to exclude from the generated code.Example: EntityType1,EntityType2,EntityType3"
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-function-imports|-efi"},
+                        "Comma-separated list of the names of function imports to exclude from the generated code.Example: FunctionImport1,FunctionImport2"
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-action-imports|-eai"},
+                        "Comma-separated list of the names of action imports to exclude from the generated code.Example: ActionImport1,ActionImport2"
+                    ),
+                    new Option<bool?>(
+                        new[] { "--ignore-unexpected-elements|-iue"},
+                        "This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate the client code if any."
+                    ),
+                    new Option<string>(
+                        new[] { "--file-name|-fn"},
+                        "The file name of the generated proxy classes."
+                    ),
+                    new Option<string>(
+                        new[] { "--outputdir|-o"},
+                        "Full path to output directory."
+                    )
+                }
+            };
 
-            app.Name = "odata-cli";
-            app.Description = "Client Proxy generator for OData endpoints.";
-
-            // Set the arguments to display the description and help text
-            app.HelpOption("-?|-h|--help");
+            await app.InvokeAsync(args);
         }
     }
 }


src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                new Command("generate", "Command to generate proxy classes for OData endpoints.")
+                {
+                    new Option<string>(
+                        new[] { "--metadata", "-m" },
+                        "The URI of the metadata document. The value must be set to a valid service document URI or a local file path."
+                    ){IsRequired = true},
+                    new Option<string>(
+                        new[] { "--custom-headers|--h"},
+                        "Headers that will get sent along with the request when fetching the metadata document from the service."
+                    ),
```
habbes, line 22: What format will the headers be provided in?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    new Option<string>(
+                        new[] { "--proxy|--p"},
+                        "Proxy settings. Format: domain\\user:password@SERVER:PORT."
+                    ),
+                    new Option<string>(
+                        new[] { "--namespace|--ns"},
+                        "The namespace of the client code generated."
```
habbes, line 29: Maybe provide an example here as well?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                        new[] { "--no-tracking|-n"},
+                        "The use of this disables entity and property tracking."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-alias|-na"},
+                        "This flag indicates whether to disable naming alias."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--internal|-i"},
+                        "If set to true, generated types will have an \"internal\" class modifier instead of \"public\" thereby making them invisible outside the assembly."
```
habbes, line 41: I assume the user won't set this to true per se, but rather they would just provide the option. So maybe it could be reworded? e.g. `Apply the "internal" class modifier on generated classes instead of "public" thereby making them invisible outside the assembly.`

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                        new[] { "--namespace|--ns"},
+                        "The namespace of the client code generated."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-tracking|-n"},
+                        "The use of this disables entity and property tracking."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-alias|-na"},
+                        "This flag indicates whether to disable naming alias."
```
habbes, line 37: It's not clear what this option means. Could you provide an example?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                        new[] { "--proxy|--p"},
+                        "Proxy settings. Format: domain\\user:password@SERVER:PORT."
+                    ),
+                    new Option<string>(
+                        new[] { "--namespace|--ns"},
+                        "The namespace of the client code generated."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-tracking|-n"},
+                        "The use of this disables entity and property tracking."
```
habbes, line 33:
```suggestion
                        "Disable entity and property tracking."
```
I think you should add more info to explain what the impact of disabling entity/property tracking is.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                        new[] { "--no-alias|-na"},
+                        "This flag indicates whether to disable naming alias."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--internal|-i"},
+                        "If set to true, generated types will have an \"internal\" class modifier instead of \"public\" thereby making them invisible outside the assembly."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--multiple-files"},
+                        "This files indicates whether to generate the files into multiple files or single."
```
habbes, line 45:
```suggestion
                        "Split the generated classes into separate files instead of generating all the code in a single file."
```

habbes: Maybe you should also have a short-form alias for this option?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-eoi"},
+                        "Comma-separated list of the names of operation imports to exclude from the generated code."
```
habbes, line 49: I think you should have a full-form alias for this option, e.g. `--exclude-operation-imports`. And maybe also provide an example in the description.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-ebo"},
+                        "Comma-separated list of the names of bound operations to exclude from the generated code."
```
habbes, line 53: Consider having a long-form option name and an example here as well.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-est"},
+                        "Comma-separated list of the names of entity types to exclude from the generated code."
```
habbes, line 57: Consider having a long-form option name and an example here as well.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-efi"},
+                        "Comma-separated list of the names of function imports to exclude from the generated code."
```
habbes, line 61: Consider having a long-form option name and an example here as well.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-eai"},
+                        "Comma-separated list of the names of action imports to exclude from the generated code."
```
habbes, line 65: Consider having a long-form option name and an example here as well.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "-iue"},
+                        "This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate the client code if any."
```
habbes, line 69: Consider having a long-form option name and an example here as well.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+                    )
+                }
+            };
 
-            app.Name = "odata-cli";
-            app.Description = "Client Proxy generator for OData endpoints.";
-
-            // Set the arguments to display the description and help text
-            app.HelpOption("-?|-h|--help");
```
habbes, line 15: Why did you remove these?

src/OData.Cli/OData.Cli.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.1.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21308.1" />
```
gathogojr, line 11: Should we be referencing a beta version?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,77 @@
...
+            var app = new RootCommand
+            {
+                new Command("generate", "Command to generate proxy classes for OData endpoints.")
+                {
+                    new Option<string>(
+                        new[] { "--metadata", "-m" },
+                        "The URI of the metadata document. The value must be set to a valid service document URI or a local file path."
+                    ){IsRequired = true},
+                    new Option<string>(
+                        new[] { "--custom-headers|--h"},
```
gathogojr, line 20: How come the shortened option starts with `-` in some cases and `--` in some other cases? E.g. `-m` vs `--h`

src/OData.Cli/Program.cs
gathogojr: `excluded-entity-types` feels like a good choice

src/OData.Cli/Program.cs
gathogojr: `ignore-unexpected-elements` feels like a good choice

gathogojr:
@ElizabethOkerio A few points to consider:
1. How does someone specify service name?
2. Should we not allow someone the option of specifying a settings config file to be used?
3. Should we not allow someone the option of specifying a file location for the metadata file, i.e., not just an web address?
4. Should we not allow someone the option of saving the settings to a config file?
5. Should we not allow someone the option of specifying the name of the proxy classes file?
6. Should we not allow someone the option of dictating whether C# casing style should be applied?
7. Should we not allow someone the option of emitting T4 files? For example, if they choose to use Visual Studio after

ElizabethOkerio:
> @ElizabethOkerio A few points to consider:
> 
> 1. How does someone specify service name?
> 2. Should we not allow someone the option of specifying a settings config file to be used?
> 3. Should we not allow someone the option of specifying a file location for the metadata file, i.e., not just a web address?
> 4. Should we not allow someone the option of saving the settings to a config file?
> 5. Should we not allow someone the option of specifying the name of the proxy classes file?
> 6. Should we not allow someone the option of dictating whether C# casing style should be applied?
> 7. Should we not allow someone the option of emitting T4 files? For example, if they choose to use Visual Studio after

@gathogojr  I don't think we'll need a service name - this was mostly for the connected service. I could be wrong though. For the settings config file also - Am not sure we need it. This tool will be generating only the proxy classes and the metadata file - I don't think it will have the ConnectedService.json file. I didn't see the need for the settings file.  With the metadata command, you can provide a location to a file or a web address.

src/OData.Cli/OData.Cli.csproj
ElizabethOkerio: The latest release for this package is in beta. The previous release was alpha.

src/OData.Cli/Program.cs
ElizabethOkerio: have set it as a bool - they should pass true or false

gathogojr:
> > @ElizabethOkerio A few points to consider:
> > 
> > 1. How does someone specify service name?
> > 2. Should we not allow someone the option of specifying a settings config file to be used?
> > 3. Should we not allow someone the option of specifying a file location for the metadata file, i.e., not just a web address?
> > 4. Should we not allow someone the option of saving the settings to a config file?
> > 5. Should we not allow someone the option of specifying the name of the proxy classes file?
> > 6. Should we not allow someone the option of dictating whether C# casing style should be applied?
> > 7. Should we not allow someone the option of emitting T4 files? For example, if they choose to use Visual Studio after
> 
> @gathogojr I don't think we'll need a service name - this was mostly for the connected service. I could be wrong though. For the settings config file also - Am not sure we need it. This tool will be generating only the proxy classes and the metadata file - I don't think it will have the ConnectedService.json file. I didn't see the need for the settings file. With the metadata command, you can provide a location to a file or a web address.

@ElizabethOkerio Thank you for you response. You make valid points. The thing I have a different view about is the config file. Especially because of things like excluded schema types, operation imports, functions and actions that would require the user to do a lot of types, the config file would come in handy since it'd make it many times easier for the user to use the tools subsequent times

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,93 @@
...
+        static async Task Main(string[] args)
         {
-            var app = new CommandLineApplication();
-
-            app.Name = "odata-cli";
-            app.Description = "Client Proxy generator for OData endpoints.";
-
-            // Set the arguments to display the description and help text
-            app.HelpOption("-?|-h|--help");
+            System.Console.WriteLine("popos");
```
KenitoInc, line 10: Why do we have this line?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,93 @@
...
+            var app = new RootCommand
```
KenitoInc, line 11: We should have a `--help` option

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,91 @@
...
+                    new Option<string>(
+                        new[] { "--file-name|-fn"},
+                        "The file name of the generated proxy classes."
+                    ),
+                    new Option<bool>(
+                        new[] { "--c#-casing-style|-ccs"},
+                        "This flag indicates whether to use the C# casing style or not."
+                    ),
+                    new Option<bool>(
+                        new[] { "--include-t4-files|-itt"},
```
gathogojr, line 83: @ElizabethOkerio Can you get the opinion of others whether we need to provide the option of emitting T4 files when someone is using this tool? Alternatively we could leave out the option and add it later if it's requested

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,91 @@
...
+                        new[] { "--namespace|-ns"},
+                        "The namespace of the client code generated. Example:ODataCliCodeGeneratorSample.NorthWindModel or ODataCliCodeGeneratorSample or it could be a name related to the OData endpoint."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-tracking|-n"},
+                        "Disables entity and property tracking. Entity tracking enables the client to track an entity in DataServiceContext while property tracking enables the client to track only top-level properties."
+                    ),
+                    new Option<bool?>(
+                        new[] { "--no-alias|-na"},
+                        "Disables naming alias."
```
habbes, line 40: It's not clear what this option means. What is "naming alias"?

habbes: I think the option is used to enable PascalCase (i.e. `alertType` would be renamed to `AlertType` when this option is enabled).

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,91 @@
...
+                    new Option<bool>(
+                        new[] { "--ignore-unexpected-elements|-iue"},
+                        "This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate the client code if any."
+                    ),
+                    new Option<string>(
+                        new[] { "--file-name|-fn"},
+                        "The file name of the generated proxy classes."
+                    ),
+                    new Option<bool>(
+                        new[] { "--c#-casing-style|-ccs"},
```
habbes, line 79: I think `csharp-casing` or `cs-casing` instead of `c#` to avoid the special character? Not sure it's a big deal though.

src/OData.Cli/Program.cs
habbes: I see there's another option for c#/PascalCasing. So I'm still unsure what this option does.

habbes (approved):
Minor comments. But looks okay.

src/OData.Cli/Program.cs
ElizabethOkerio: disables upper camel casing. it is enabled by default.

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,87 @@
...
+            {
+                new Command("generate", "Command to generate proxy classes for OData endpoints.")
+                {
+                    new Option<string>(
+                        new[] { "--metadata|-m"},
+                        "The URI of the metadata document. The value must be set to a valid service document URI or a local file path."
+                    ){IsRequired = true},
+                    new Option<string>(
+                        new[] { "--load-settings-from-file|-lsff" },
+                        "The URI to the settings file."
```
gathogojr, line 20:
```suggestion
                        "The location of the configuration settings file."
```
Because URI suggests a web location

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,87 @@
...
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-function-imports|-efi"},
+                        "Comma-separated list of the names of function imports to exclude from the generated code.Example: FunctionImport1,FunctionImport2"
+                    ),
+                    new Option<string>(
+                        new[] { "--excluded-action-imports|-eai"},
+                        "Comma-separated list of the names of action imports to exclude from the generated code.Example: ActionImport1,ActionImport2"
+                    ),
+                    new Option<bool>(
```
gathogojr, line 70: Should this be `new Option<bool?>(` as well?

src/OData.Cli/Program.cs
```diff
@@ -1,18 +1,87 @@
...
+                    new Option<string>(
+                        new[] { "--file-name|-fn"},
+                        "The file name of the generated proxy classes."
+                    ),
+                    new Option<string>(
+                        new[] { "--outputdir|-o"},
+                        "Full path to output directory."
+                    )
+                }
+            };
```
gathogojr, line 83:
```suggestion
            };

```

gathogojr (approved):
Left some comments otherwise looks good

PR #238, opened 2021-12-09
move core classes to ODataCodeGen.Core

ElizabethOkerio:
In this PR, I have moved all the core files to the <code> ODataCodeGen.Core </code> class library. No major change has been made. It is just moving the classes that will be shared by both <code>ODataConnectedService </code> and <code>OData.Cli </code> projects. In the next PR, I will decouple the visual studio dependencies in some of these classes.


habbes (approved):

habbes:
Suggestion for the root namespace for the new project: `Microsoft.OData.CodeGen`.

gathogojr (approved):
+1 to the naming suggestion by @habbes; otherwise looks good

PR #178, opened 2020-11-20
Fix #176 : Create new Service Proxy or regenerate existing Service Proxy fails (OData V3, WCF Data Services)

unchase:
Fix #176.
There are a few errors in v0.11.1:

issue176

issue176_1

issue176_2

for metadata:

```xml
<?xml version="1.0" standalone="yes" ?>
<edmx:Edmx Version="1.0" xmlns:edmx="http://schemas.microsoft.com/ado/2006/04/edmx">
  <edmx:DataServices xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata" m:DataServiceVersion="2.0">
    <Schema Namespace="Simple" xmlns="http://schemas.microsoft.com/ado/2006/04/edmx">
      <EntityType Name="TestType">
        <Key>
          <PropertyRef Name="KeyProp" />
        </Key>
        <Property Name="KeyProp" Type="Edm.Int32" Nullable="false" />
        <Property Name="ValueProp" Type="Edm.String" Nullable="false" />
      </EntityType>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx> 
```

fixed.

---

Added the tests below:

To **EdmHelperTests:**

- [x] `TestGetSchemaTypesWithNullModel`;
- [x] `TestGetOperationImportsWithNullModel`;
- [x] `TestGetBoundOperationsWithNullModel`.

To **CodeGenDescriptorTest:**

- [x] `TestV2AddGeneratedClientCode_GeneratesCodeForv2_ForVB`;
- [x] `TestV2AddGeneratedClientCode_GeneratesCodeForv2`.

---

Also improved logging the *'Could not get CLR type name for EDM type'* exception in OData v4 code generation from:

edmTypeError

to:

edmTypeErrorDetalization

for metadata (for example): 

```xml
<?xml version="1.0" standalone="yes" ?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
    <Schema Namespace="Simple" xmlns:d="http://docs.oasis-open.org/odata/ns/data" 
            xmlns:m="http://docs.oasis-open.org/odata/ns/metadata" 
            xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityType Name="TestType">
        <Key>
          <PropertyRef Name="KeyProp" />
        </Key>
        <Property Name="KeyProp" Type="Edm.Int32" Nullable="false" />
[...]

Issue #176, opened 2020-11-10
Create new Service Proxy or regenerate existing Service Proxy fails (OData V3, WCF Data Services)


algolsoftware:
Generating a new client proxy fails with "GetMSBuildProjectForHierarchy must be called on the UI thread".

ODataConnectedService v0.11.1
VisualStudio Version v16.7.7
Microsoft.Data.Odata v5.8.4
Microsoft.Data.Services.Client v5.8.4
Microsoft.Data.Services v5.8.4

1. Create WebApi project with simple code-first EF context and WCF Data Service for this context.
2. Create MVC project for client
3. Browse the Data Service Url (VS show in browser)
4. In the client project add connected service / ODataConnectedService
5. Set Address to the service Url
6. Click Finish

The progess dialog shows the error:
![image](https://user-images.githubusercontent.com/49318148/98658300-2d9d5a00-2343-11eb-9bfb-f6bfacff6cd6.png)

Clicking on the Next button after step 5 (to Schema Types Selection) the UI stops reacting on the wizard buttons except the Cancel button.
Clicking on Cancel button or closing the wizard window closes the window.
Clicking on the Finish button first and then on the Cancel button or closing the wizard window results in the following error:
![image](https://user-images.githubusercontent.com/49318148/98659022-01360d80-2344-11eb-8bc0-229b55cdd143.png)
- looks like Dictionary Get with NULL key...

This is the same behavior when trying to regenerate an existing service proxy:
1. Select Refresh Connected Service on existing service
2. Click finish in the Wizard window

Results in the error in the last screenshot. 
The behavior after clicking the Next button ist the same as decribed in the case of a new service proxy.  

Update: 
ODataConnectedService v0.10.0 works with one exception:
Refreshing an existing service proxy: clicking on Finish without first clicking Next causes the same error.

tfaller:
I can confirm this (with OData service based of Dynamics Nav 2016).
The problem is not there at OData Connected Service version 0.10.0.

doug-coates:
I'm having the same issue with VS 2019...

unchase:
@algolsoftware @tfaller @doug-coates can you provide your xml metadata?

doug-coates:
Here’s my metadata…

unchase:
@doug-coates I can't see the attached file.
Can you send the file to spiritkola@hotmail.com (directly from email)?
Or add xml-metadata in code-block like this:

```xml
<?xml version="1.0" standalone="yes" ?>
[...]

PR #241, opened 2022-01-27
decouple odata.core classes from visual studio

ElizabethOkerio:
This PR decouples Visual Studio from the <code>Microsoft.OData.CodeGen </code> library. 

I have added 3 interfaces 
- _IFileHandler_
- _IPackageInstaller_
- _IMessageLogger_

These interfaces have some wrapper methods that can be implemented differently depending on who is implementing them.


src/Microsoft.OData.CodeGen/CodeGeneration/BaseCodeGenDescriptor.cs
```diff
@@ -35,60 +31,25 @@ public abstract class BaseCodeGenDescriptor
 
         protected string CurrentAssemblyPath => Path.GetDirectoryName(this.GetType().Assembly.Location);
         
-        protected LanguageOption TargetProjectLanguage => this.Project.GetLanguageOption();
+        protected LanguageOption TargetProjectLanguage { get; set; }
 
-        protected BaseCodeGenDescriptor(string metadataUri, ConnectedServiceHandlerContext context, Project project)
+        protected BaseCodeGenDescriptor(string metadataUri, string filePath, LanguageOption languageOption, IFileHandler fileHandler,IMessageLogger logger, ServiceConfiguration serviceConfiguration, IPackageInstaller packageInstaller)
```
habbes, line 36:
```suggestion
        protected BaseCodeGenDescriptor(string metadataUri, string filePath, LanguageOption languageOption, IFileHandler fileHandler, IMessageLogger logger, ServiceConfiguration serviceConfiguration, IPackageInstaller packageInstaller)
```

src/Microsoft.OData.CodeGen/CodeGeneration/BaseCodeGenDescriptor.cs
```diff
@@ -5,28 +5,24 @@
...
-        public IVsPackageInstaller PackageInstaller { get; protected set; }
-        public IVsPackageInstallerServices PackageInstallerServices { get; protected set; }
-        public ConnectedServiceHandlerContext Context { get; private set; }
+        public IPackageInstaller PackageInstaller { get; protected set; }
+        public IFileHandler FileHandler { get; private set; }
+        public IMessageLogger MessageLogger { get; private set; }
         public ServiceConfiguration ServiceConfiguration { get; set; }
-        public Project Project { get; private set; }
         public string MetadataUri { get; private set; }
+        public string FilePath { get; private set; }
```
habbes, line 25: This actually refers to a folder path right? Maybe a better name would be something like `destinationPath` or `outputDirectory`

src/Microsoft.OData.CodeGen/Logging/IMessageLogger.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.OData.CodeGen.Logging
+{
+    /// <summary>
+    /// Message logging interface
+    /// </summary>
+    public interface IMessageLogger
+    {
+        Task WriteMessageAsync(LogMessageCategory logMessageCategory, string message, params object[] args);
```
habbes, line 17: Add doc comments

src/Microsoft.OData.CodeGen/PackageInstallation/IPackageInstaller.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.OData.CodeGen.PackageInstallation
+{
+    /// <summary>
+    /// Package installation interface
+    /// </summary>
+    public interface IPackageInstaller
+    {
+        Task CheckAndInstallNuGetPackageAsync(string packageSource, string packageName);
```
habbes, line 17: Add doc comments

src/Microsoft.OData.CodeGen/CodeGeneration/BaseCodeGenDescriptor.cs
habbes: I think instead of passing the filePath to the constructor and storing it as a property, it would make more sense to pass it to the method that generates the files. If you have multiple destination files (hypothetically), it you wouldn't need to create multiple instances of the code gen descriptor.

src/Microsoft.OData.CodeGen/CodeGeneration/V3CodeGenDescriptor.cs
```diff
@@ -42,7 +44,7 @@ public override async Task AddGeneratedClientCodeAsync()
 
         private async Task AddGeneratedCodeAsync()
         {
-            await Context.Logger.WriteMessageAsync(LoggerMessageCategory.Information, "Generating Client Proxy ...").ConfigureAwait(true);
+            await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Generating Client Proxy ...").ConfigureAwait(true);
```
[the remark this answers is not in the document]
ElizabethOkerio: From the discussions I had with Clement, It seemed like the ConfigureAwait(true) was added to fix this issue https://github.com/OData/ODataConnectedService/issues/176 which occurred when some methods were being accessed outside of the UI thread. This is the PR that added this: https://github.com/OData/ODataConnectedService/pull/178/files

src/ODataConnectedService/DefaultFileHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public void AddCsdlFileAsEmbeddedResource(string fileName)
+        {
+            var dte = Microsoft.VisualStudio.Shell.Package.GetGlobalService(typeof(DTE)) as DTE;
+            if (dte != null)
+            {
+                var projectItem = this.Project.ProjectItems.Item("Connected Services").ProjectItems.Item(((ODataConnectedServiceInstance)this.Context.ServiceInstance).ServiceConfig.ServiceName).ProjectItems.Item(fileName);
+                projectItem.Properties.Item("BuildAction").Value = prjBuildAction.prjBuildActionEmbeddedResource;
+            }
+        }
```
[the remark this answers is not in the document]
ElizabethOkerio: We'll cross the bridge when we get there.

src/ODataConnectedService/ConnectedServiceMessageLogger.cs
```diff
@@ -0,0 +1,27 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="DefaultMessageLogger.cs" company=".NET Foundation">
```
KenitoInc, line 2:
```suggestion
// <copyright file="ConnectedServiceMessageLogger.cs" company=".NET Foundation">
```

src/ODataConnectedService/ConnectedServicePackageInstaller.cs
```diff
@@ -0,0 +1,75 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="DefaultPackageInstaller.cs" company=".NET Foundation">
```
KenitoInc, line 2:
```suggestion
// <copyright file="ConnectedServicePackageInstaller.cs" company=".NET Foundation">
```

test/ODataConnectedService.Tests/CodeGeneration/CodeGenDescriptorTest.cs
```diff
@@ -25,6 +25,10 @@
 using ODataConnectedService.Tests;
 using ODataConnectedService.Tests.TestHelpers;
 using Common = Microsoft.OData.CodeGen.Common;
+using Microsoft.OData.CodeGen.FileHandling;
+using Microsoft.OData.CodeGen.Logging;
+using Microsoft.OData.CodeGen.PackageInstallation;
+using System.Data.Services.Design;
```
KenitoInc, line 31: re-arrange these usings

KenitoInc (commented):
A few comments

src/Microsoft.OData.CodeGen/FileHandling/ODataFileOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// </summary>
+        public ODataFileOptions()
+        { 
+        }
+
+        //
+        // Summary:
+        //     Gets or sets a value indicating whether to suppress prompting the end user if
+        //     an existing file is detected in the target path and should be overwritten. The default is false
+        public bool SuppressOverwritePrompt { get; set; }
```
KenitoInc, line 26:
```suggestion
        public bool SuppressOverwritePrompt { get; set; }
        
```

src/Microsoft.OData.CodeGen/FileHandling/ODataFileOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Instantiates a new instance of the ODataFileOptions class.
+        /// </summary>
+        public ODataFileOptions()
+        { 
+        }
+
+        //
+        // Summary:
```
KenitoInc, line 23: Add a docstring in the correct format
```csharp
/// <summary>
/// Gets or sets a value indicating whether to suppress prompting the end user if
/// an existing file is detected in the target path and should be overwritten. The default is false
/// </summary>
```

src/Microsoft.OData.CodeGen/FileHandling/ODataFileOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+        //     Gets or sets a value indicating whether to suppress prompting the end user if
+        //     an existing file is detected in the target path and should be overwritten. The default is false
+        public bool SuppressOverwritePrompt { get; set; }
+        //
+        // Summary:
```
KenitoInc, line 28: Add docstring in the correct format

src/ODataConnectedService/ConnectedServiceMessageLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+// </copyright>
+//----------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using Microsoft.OData.CodeGen.Logging;
+using Microsoft.VisualStudio.ConnectedServices;
+
+namespace Microsoft.OData.ConnectedService
+{
+    public class ConnectedServiceMessageLogger : IMessageLogger
```
KenitoInc, line 14: Add docstring in the public class and methods

src/ODataConnectedService/ConnectedServicePackageInstaller.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.OData.CodeGen.Logging;
+using Microsoft.OData.CodeGen.PackageInstallation;
+using Microsoft.VisualStudio.ComponentModelHost;
+using Microsoft.VisualStudio.ConnectedServices;
+using NuGet.VisualStudio;
+using Shell = Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.OData.ConnectedService
+{
+    public class ConnectedServicePackageInstaller : IPackageInstaller
```
KenitoInc, line 20: Add docstring in the public class and methods

KenitoInc (approved):
LGTM

src/Microsoft.OData.CodeGen/CodeGeneration/V3CodeGenDescriptor.cs
```diff
@@ -11,28 +11,30 @@
...
+using Microsoft.OData.CodeGen.Models;
+using Microsoft.OData.CodeGen.PackageInstallation;
 
 namespace Microsoft.OData.CodeGen.CodeGeneration
 {
     public class V3CodeGenDescriptor : BaseCodeGenDescriptor
     {
-        public V3CodeGenDescriptor(string metadataUri, ConnectedServiceHandlerContext context, Project project)
-            : base(metadataUri, context, project)
+        public V3CodeGenDescriptor(string metadataUri,string outputDirectory, LanguageOption languageOption, IFileHandler fileHandler, IMessageLogger logger, ServiceConfiguration serviceConfiguration, IPackageInstaller packageInstaller)
```
habbes, line 23:
```suggestion
        public V3CodeGenDescriptor(string metadataUri, string outputDirectory, LanguageOption languageOption, IFileHandler fileHandler, IMessageLogger logger, ServiceConfiguration serviceConfiguration, IPackageInstaller packageInstaller)
```

habbes: I think some of these arguments, like `outputDirectory` and maybe others should not be constructor arguments or properties of the code generator but rather arguments passed to the method that actually does code generation (i.e. `AddGeneratedClientCodeAsync`). With the way it currently setup, if we had to generate code multiple times for different inputs, we would need to create a new instance of the code generator each time even though the behaviour of the different instances would be exactly the same, only the output would change. I think this makes the code generator unnecessarily stateful.

src/Microsoft.OData.CodeGen/CodeGeneration/V3CodeGenDescriptor.cs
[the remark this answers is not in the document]
habbes: Instead of `ConfigureAwait(true)`, we should just remove that call since `ConfigureAwait(true)` is the default. I think the only reason we added `ConfigureAwait(false)` in the first place was to satisfy static analyzer suggestions which were not aware of the context of this app. But we still have cases where we use `ConfigureAwait(true)` and other places we use `ConfigureAwait(false)`.

habbes (approved):
Looks good. Left some suggestions, but those could also be revisited in future iterations to avoid blocking this PR.

gathogojr (approved):

PR #239, opened 2021-12-12
feat(VisualStudio2022): Update the Plugin to VS2022 Version for x64

felix1000:
It is a very quick update to let this extension can run on VS2022.
Take if it is useful

PR #274, opened 2022-04-10
feat(VisualStudio2022): Update the Plugin to VS2022 Version for x64

1Jesper1:
Fixed review of https://github.com/OData/ODataConnectedService/pull/239 and merged main branch

Got some problems building Microsoft.Odata.CodeGen project, because of missing Microsoft.Odata references.
Got a problem building ODataConnectedService, because of an error Json.Newtonsoft property.
Fixed these locally, but did no commit these.
Build fails see https://dev.azure.com/dotnet/OData/_build/results?buildId=69910&view=results, I'm going to commit the local fixes => https://github.com/1Jesper1/ODataConnectedService-For-VS2022/commit/cee3e15d4e8c2c531da708184d41f31caed23ce2


1Jesper1:
@KenitoInc @habbes @henning-krause Please take a look

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
```diff
@@ -61,15 +61,15 @@
       <HintPath>..\..\external\Binaries\V3\Microsoft.Data.Services.Design.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.OData.Client, Version=7.6.3.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.OData.Client.7.6.3\lib\net45\Microsoft.OData.Client.dll</HintPath>
+      <HintPath>packages\Microsoft.OData.Client.7.6.3\lib\net45\Microsoft.OData.Client.dll</HintPath>
```
ElizabethOkerio, line 64: why are you updating this path? is this the right path?

ElizabethOkerio: I though don't think it is necessary to add the packages.config file in every project. We can just use the one in the root folder.

src/ODataConnectedService/ODataConnectedService.csproj
```diff
@@ -150,20 +145,12 @@
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
     </Reference>
     <Reference Include="netstandard" />
-    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\..\RepoConnectedService\ODataConnectedService\packages\Newtonsoft.Json.12.0.3\lib\net45\Newtonsoft.Json.dll</HintPath>
-    </Reference>
     <Reference Include="NuGet.VisualStudio, Version=4.0.0.2323, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\..\packages\NuGet.VisualStudio.4.0.0\lib\net45\NuGet.VisualStudio.dll</HintPath>
```
ElizabethOkerio, line 149: Why didn't you make the path changes to all references then?

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
1Jesper1: I also think, but I can't get it working with root.

src/ODataConnectedService/ODataConnectedService.csproj
1Jesper1: I got a Json.Newtonsoft error, error said it was because two different Newtonsoft.Json packages were present.

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
ElizabethOkerio: which issue are you facing? I could assist.

1Jesper1: https://dev.azure.com/dotnet/OData/_build/results?buildId=69910&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=98480a0e-3db2-53e6-5e37-d90e8e661f3c this was the error without these changes.

1Jesper1: Btw, can we delete OData.CLI.sln?

pbvs:
Hi @1Jesper1 , i have conducted some tests as you requested in issue #239. 

Please see my findings below:
- Build from of vsix package from VS2022 : I am able to create a vsix package without any issues
- The commandline build is still tied to vs2017 and does not work
- Installation of the vsix package proceeds without any issues. After installation the Odata Connected Service is available!! Very Happy with this :)
image
- I am able to generate two odata connected services based on different odata metadata files
image
- I noticed 2 differences between VS2019 and VS2022 is the option "Update OData Connected Service" is missing in vs2022 and the option manage connected services is missing in vs2022
image
image

I will continue testing the coming days and if i find anything else if will update this thread.

1Jesper1:
@pbvs Thanks for testing! Good to hear you see the Odata service available now! I will look into the commandline build problem and the 2 differences.

pbvs:
Hi @1Jesper1 i am not yet 100% sure, but it looks like this version does not work in VS2019 and VX2017. When generating a service context i get the following error
image

Unfortunately i cant see exactly which dll can't be found. I will now try the following:
- Uninstall All odata vsix files (all versions of VS)
- Manually check the folder Enterprise\Common7\IDE\Extensions that all files are gone, if not manually delete
- Restart machine
- install original version in VS2019 and check if i can generate service client
- Repeat above steps
- Install new version from #274 and check if i can generate service client in vs2019

I will let you know the results.

1Jesper1:
@pbvs How did you test the command line build?

1Jesper1:
@pbvs It seems https://github.com/OData/ODataConnectedService/pull/239#issuecomment-1041346701 Odata service is not visible, thought I checked it, but it seems it's still a problem when installing the extension instead of starting debugging.

1Jesper1:
@pbvs Command line build command is fixed, see https://github.com/OData/ODataConnectedService/pull/274/commits/662843ffe12a8f7e2ba205723a43ad013f7e8be6 but you need to comment the 2017 and 2019 locations in build.cmd.

1Jesper1:
@pbvs Installing extension in 2017 and 2019 does not work, we need to start using shared project and different vsix projects. See https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/target-previous-versions?view=vs-2022

1Jesper1:
@pbvs Very weird indeed that "Manage connected services" and "Update Odata connected service" is not available.
Workaround => Right click on project => Add => Connected service.

1Jesper1:
https://github.com/OData/ODataConnectedService/pull/295 merged, this is obsolete and can be closed.

PR #295, opened 2022-06-02
support vs2022

ElizabethOkerio:
This PR follows this guideline: https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022. in updating the OData Connected Service extension for visual studio 2022.


src/ODataConnectedService_vs2022/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+// Version information for an assembly consists of the following four values:
+//
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
```
gathogojr, line 31:
```suggestion
```

src/ODataConnectedService/source.extension.vsixmanifest
```diff
@@ -1,13 +1,13 @@
...
     <Metadata>
-        <Identity Id="ODataConnectedService.e30335d6-f9c7-4d08-b66a-f011f3f18477" Version="0.13.0" Language="en-US" Publisher="Microsoft" />
+        <Identity Id="ODataConnectedService.e30335d6-f9c7-4d08-b66a-f011f3f18477" Version="0.13.3" Language="en-US" Publisher="Microsoft" />
         <Author>Microsoft</Author>
         <DisplayName>OData Connected Service</DisplayName>
         <Description xml:space="preserve">OData Connected Service for V1-V4</Description>
         <MoreInfo>https://github.com/odata/ODataConnectedService</MoreInfo>
         <License>License.txt</License>
-        <Icon>Resources\ExtensionIcon.png</Icon>
+        <Icon>Resources\Icon.png</Icon>
```
gathogojr, line 10: Why?

habbes:
Some initial thoughts:

We seem to have 3 sln files and some of them seem redundant.

The `ODataConnectedService.Tests.sln` contains all the projects. This makes the other sln files feel redundant. Maybe we can keep this one solution and remove the rest. And also rename it from `ODataConnectedService.Tests.sln` to something else. Since now it contains more than just the `ODataConnectedService` project (i.e. it also contains the CLI and the CodeGen lib), maybe it can be renamed to something like `ODataTools.sln`.

If the aim was to separate the CLI project from the connected service, we could have 2 solution files `OData.Cli.sln` and `ODataConnectedService.sln`. However, I'm not sure if this is more beneficial than just having a single solution, especially since they would all reference the CodeGen common project.

The new project for VS 2022 is called `ODataConnectedService_vs2022`. Since this is probably what's going to the "current" version, I suggest we rename it to `ODataConnectedService` and then rename the project targeting VS 2019 and lower to something like `ODataConnectedService_Legacy`.

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
```diff
@@ -193,10 +124,38 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="OcsKey.snk" />
-    <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces">
```
KenitoInc, line 132: What's the criteria behind choosing which packages to move to `PackageReference`?

src/ODataConnectedService/source.extension.vsixmanifest
KenitoInc: The 2 images look the same, with same size

habbes:
Are there any steps I need to take to get this to run locally? I have tried to debug the `ODataConnectedService_vs2022` project after opening the solution in VS 2022, and it fails. I also tried `dotnet restore` and it also fails.

VsBuildError


MicrosoftTeams-image (1)

MicrosoftTeams-image (2)

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
ElizabethOkerio: all packages from nuget are referenced using the `PackageReference`

ElizabethOkerio:
> Are there any steps I need to take to get this to run locally? I have tried to debug the `ODataConnectedService_vs2022` project after opening the solution in VS 2022, and it fails. I also tried `dotnet restore` and it also fails.
> 
> VsBuildError
> 
> MicrosoftTeams-image (1)
> 
> MicrosoftTeams-image (2)

Have pulled this branch again and was able to run it. Can you try the other VSIX for visual studio 2019 and lower versions and let me know if it works.

ElizabethOkerio:
> Some initial thoughts:
> 
> We seem to have 3 sln files and some of them seem redundant.
> 
> The `ODataConnectedService.Tests.sln` contains all the projects. This makes the other sln files feel redundant. Maybe we can keep this one solution and remove the rest. And also rename it from `ODataConnectedService.Tests.sln` to something else. Since now it contains more than just the `ODataConnectedService` project (i.e. it also contains the CLI and the CodeGen lib), maybe it can be renamed to something like `ODataTools.sln`.
> 
> If the aim was to separate the CLI project from the connected service, we could have 2 solution files `OData.Cli.sln` and `ODataConnectedService.sln`. However, I'm not sure if this is more beneficial than just having a single solution, especially since they would all reference the CodeGen common project.
> 
> The new project for VS 2022 is called `ODataConnectedService_vs2022`. Since this is probably what's going to the "current" version, I suggest we rename it to `ODataConnectedService` and then rename the project targeting VS 2019 and lower to something like `ODataConnectedService.Legacy`.

From the discussions that we had, Legacy  is not the right word to use. We settled for `ODataConnectedService_Latest` for `VS 2022` project

ElizabethOkerio:
> > Some initial thoughts:
> > We seem to have 3 sln files and some of them seem redundant.
> > The `ODataConnectedService.Tests.sln` contains all the projects. This makes the other sln files feel redundant. Maybe we can keep this one solution and remove the rest. And also rename it from `ODataConnectedService.Tests.sln` to something else. Since now it contains more than just the `ODataConnectedService` project (i.e. it also contains the CLI and the CodeGen lib), maybe it can be renamed to something like `ODataTools.sln`.
> > If the aim was to separate the CLI project from the connected service, we could have 2 solution files `OData.Cli.sln` and `ODataConnectedService.sln`. However, I'm not sure if this is more beneficial than just having a single solution, especially since they would all reference the CodeGen common project.
> > The new project for VS 2022 is called `ODataConnectedService_vs2022`. Since this is probably what's going to the "current" version, I suggest we rename it to `ODataConnectedService` and then rename the project targeting VS 2019 and lower to something like `ODataConnectedService.Legacy`.
> > The project that targets earlier versions will maintain it's name `ODataConnectedService` and the project that targets `VS 2022` will be called `ODataConnectedService_Latest`

The project that targets earlier versions will maintain it's name `ODataConnectedService` and the project that targets `VS 2022` will be called `ODataConnectedService_Latest`

azure-pipelines.yml
```diff
@@ -10,17 +10,21 @@ variables:
   buildPlatform: 'Any CPU'
   buildConfiguration: 'Release'
   connectedServiceDir: '$(Build.SourcesDirectory)'
-  sln: '$(connectedServiceDir)\ODataConnectedService.sln'
-  testSln: '$(connectedServiceDir)\ODataConnectedService.Tests.sln'
+  sln: '$(connectedServiceDir)\ODataCodeGenTools.sln'
```
gathogojr, line 13: Was this change necessary?

azure-pipelines.yml
```diff
@@ -10,17 +10,21 @@ variables:
   buildPlatform: 'Any CPU'
   buildConfiguration: 'Release'
   connectedServiceDir: '$(Build.SourcesDirectory)'
-  sln: '$(connectedServiceDir)\ODataConnectedService.sln'
-  testSln: '$(connectedServiceDir)\ODataConnectedService.Tests.sln'
```
gathogojr, line 14: Why did we remove the test solution?

azure-pipelines.yml
```diff
@@ -37,29 +41,17 @@ stages:
...
 
     - task: VSBuild@1
-      displayName: 'Build solution - ODataConnectedService.sln'
+      displayName: 'Build solution - ODataCodeGenTools.sln'
       inputs:
         solution: '$(sln)'
         platform: '$(buildPlatform)'
         configuration: '$(buildConfiguration)'
 
-    - task: NuGetCommand@2
```
gathogojr, line 51: Why did we remove the tests solution?

src/ODataConnectedService_VS2022Plus/source.extension.vsixmanifest
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
+    <Metadata>
+        <Identity Id="ODataConnectedService2022plus.5b4f5787-7c18-4eb4-bf4a-5d2fcd66278a" Version="0.13.0" Language="en-US" Publisher="Company" />
+        <DisplayName>OData Connected Service 2022plus</DisplayName>
```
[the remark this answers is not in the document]
gathogojr: Please seek opinion from other contributors regarding what best to use as display name and in labels

azure-pipelines.yml
ElizabethOkerio: we didn't remove the test solution. The test sln had all the other projects. So we just named that soluton to ODataCodeGenTools.sln. which has the test project and all the other projects.

azure-pipelines.yml
```diff
@@ -10,17 +10,21 @@ variables:
...
   buildConfiguration: 'Release'
   connectedServiceDir: '$(Build.SourcesDirectory)'
-  sln: '$(connectedServiceDir)\ODataConnectedService.sln'
-  testSln: '$(connectedServiceDir)\ODataConnectedService.Tests.sln'
+  sln: '$(connectedServiceDir)\ODataCodeGenTools.sln'
   signigConfigPath: '$(connectedServiceDir)\configs'
   signingConfigFiles: '*.*'
   productBinPath: '$(connectedServiceDir)\src\ODataConnectedService\bin\$(BuildConfiguration)'
   productFiles: 'Microsoft.OData.ConnectedService.*?(*.dll|*.config|*.pdb)'
+  productBinPath1: '$(connectedServiceDir)\src\ODataConnectedService_VS2022Plus\bin\$(BuildConfiguration)'
```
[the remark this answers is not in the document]
ElizabethOkerio: resolved

src/ODataConnectedService_VS2022Plus/source.extension.vsixmanifest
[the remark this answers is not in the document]
ElizabethOkerio: fixed

src/SharedFiles/License.txt
```diff
@@ -0,0 +1,25 @@
+
```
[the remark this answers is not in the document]
ElizabethOkerio: it is different from which content? It is the same. It is the same file shared between the two projects.

src/ODataConnectedService/ODataConnectedService.csproj
```diff
@@ -237,69 +124,72 @@
     </BootstrapperPackage>
   </ItemGroup>
   <ItemGroup>
-    <Content Include="License.txt">
+    <Service Include="{508349B6-6B84-4DF5-91F0-309BEEBAD82D}" />
```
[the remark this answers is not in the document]
ElizabethOkerio: This was there before. Haven't added it.

src/SharedFiles/License.txt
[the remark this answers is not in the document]
gathogojr: Compare below with this:
image
First line on the screenshot has **MIT License (MIT)**...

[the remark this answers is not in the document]
ElizabethOkerio: seen. updated

gathogojr (approved):

habbes (changes requested):
Seems there are some things which have not been updated.

- The main `README.md` still assumes the old directory and project structure. It also assumes VS 2017 and 2019
- I tried running the `build.cmd` locally and it fails with an error message saying it couldn't copy `src/ODataConnectedService`. When I tried the command on an older branch, it succeeds. Probably needs to be updated. The `build.cmd` also references `ODataConnectedService.msbuild`, I don't know if that file also contains references that need to be updated. Frankly, I don't know if we use either of these files. Should we consider removing them?

PR #237, opened 2021-12-08
Fix #48 by improving handling of Uris with/without "$metadata" suffix.

dr-consit:
Query and fragment parts of the Uri were always clobbered, these are now
preserved, except in the special case where the last segment of the Uri is
"$metadata/".

Please note, that some of the code is perhaps slightly unwieldy, to handle
the "$metadata/" edge-case. This is done to satisfy the specific test-case:
"GetMetadataDocumentUriWithSlashShouldIgnoreQueries".

I am not sure I understand the reasoning behind this test-case, but as
mentioned, the code is made to satisfy this test.

If there is no reason to satisfy this test, that is, if it is OK to always preserve
query parameters, the whole "preserveQueryAndFragment" part of this patch
can be eliminated.

PR #302, opened 2022-07-21
Improve uri handling

### LINKED ISSUE #48: Metadata URL with values in Query String

Nthemba:
This PR fixes the issue [Metadata URL with values in Query String · Issue #48 · OData/ODataConnectedService (github.com)](https://github.com/OData/ODataConnectedService/issues/48)
and is an add on to a contributor's (https://github.com/dr-consit) PR [Fix #48 by improving handling of Uris with/without "$metadata" suffix. by dr-consit · Pull Request #237 · OData/ODataConnectedService (github.com)](https://github.com/OData/ODataConnectedService/pull/237)

It ensures the OData Endpoint URI does not get distorted when the URI includes query and fragment segments. 
### Before fix 
A URI like 'http://localhost:5000/odata/$metadata?api-version=1.0' would have $metadata appended to the end of the URI which would make the OData service inaccessible.
### After fix
$metadata, query and fragment segments maintain the correct position and order.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -101,10 +101,10 @@ stages:
         publishLocation: 'Container'
 
     - task: PublishBuildArtifacts@1
-      displayName: 'Publish Artifact - VSIX2022plus'
+      displayName: 'Publish Artifact - VSIX2022Plus'
       inputs:
-        PathtoPublish: '$(Build.ArtifactStagingDirectory)\VSIX2022plus'
-        ArtifactName: 'VSIX2022plus'
+        PathtoPublish: '$(Build.ArtifactStagingDirectory)\VSIX2022Plus'
+        ArtifactName: 'VSIX2022Plus'
         publishLocation: 'Container'
 
     - publish: configs
--- a/src/Microsoft.OData.Cli/GenerateCommand.cs
+++ b/src/Microsoft.OData.Cli/GenerateCommand.cs
@@ -138,7 +138,7 @@ namespace Microsoft.OData.Cli
                 (
                     GenerateOptions options,
                     IConsole console
-                ) 
+                )
                 => HandleGenerateCommand(options, console));
         }
 
@@ -185,10 +185,10 @@ namespace Microsoft.OData.Cli
             catch (Exception ex)
             {
                 console.Error.Write(ex.Message);
-                return 0;
+                return 1;
             }
 
-            return 1;
+            return 0;
         }
 
         private Version GetMetadataVersion(GenerateOptions generateOptions)
@@ -203,7 +203,7 @@ namespace Microsoft.OData.Cli
             serviceConfiguration.WebProxyNetworkCredentialsUsername = generateOptions.WebProxyNetworkCredentialsUsername;
             serviceConfiguration.WebProxyNetworkCredentialsPassword = generateOptions.WebProxyNetworkCredentialsPassword;
             serviceConfiguration.WebProxyNetworkCredentialsDomain = generateOptions.WebProxyNetworkCredentialsDomain;
-            
+
             version = MetadataReader.GetMetadataVersion(serviceConfiguration);
             return version;
         }
@@ -232,7 +232,7 @@ namespace Microsoft.OData.Cli
 
             Project project = ProjectHelper.CreateProjectInstance(generateOptions.OutputDir);
             BaseCodeGenDescriptor codeGenDescriptor = new CodeGenDescriptorFactory().Create(
-                Constants.EdmxVersion4, 
+                Constants.EdmxVersion4,
                 new ODataCliFileHandler(new ODataCliMessageLogger(console), project),
                 new ODataCliMessageLogger(console),
                 new ODataCliPackageInstaller(project, new ODataCliMessageLogger(console)));
--- a/src/Microsoft.OData.Cli/GenerateCommand.cs
+++ b/src/Microsoft.OData.Cli/GenerateCommand.cs
@@ -172,7 +172,8 @@ namespace Microsoft.OData.Cli
                     }
                 }
 
-                Version version = GetMetadataVersion(options);
+                ServiceConfiguration config = GetServiceConfiguration(options);
+                MetadataReader.GetMetadataVersion(config, out Version version);
                 if (version == Constants.EdmxVersion4)
                 {
                     await GenerateCodeForV4Clients(options, console).ConfigureAwait(false);
@@ -191,10 +192,9 @@ namespace Microsoft.OData.Cli
             return 0;
         }
 
-        private Version GetMetadataVersion(GenerateOptions generateOptions)
+        private ServiceConfiguration GetServiceConfiguration(GenerateOptions generateOptions)
         {
-            Version version = null;
-            var serviceConfiguration = new ServiceConfiguration();
+            ServiceConfiguration serviceConfiguration = new ServiceConfiguration();
             serviceConfiguration.Endpoint = generateOptions.MetadataUri;
             serviceConfiguration.CustomHttpHeaders = generateOptions.CustomHeaders;
             serviceConfiguration.WebProxyHost = generateOptions.WebProxyHost;
@@ -204,8 +204,7 @@ namespace Microsoft.OData.Cli
             serviceConfiguration.WebProxyNetworkCredentialsPassword = generateOptions.WebProxyNetworkCredentialsPassword;
             serviceConfiguration.WebProxyNetworkCredentialsDomain = generateOptions.WebProxyNetworkCredentialsDomain;
 
-            version = MetadataReader.GetMetadataVersion(serviceConfiguration);
-            return version;
+            return serviceConfiguration;
         }
 
         private async Task GenerateCodeForV4Clients(GenerateOptions generateOptions, IConsole console)
--- a/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
+++ b/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
@@ -36,10 +36,15 @@ namespace Microsoft.OData.CodeGen.Common
             if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!serviceConfiguration.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out var uri))
+
                 {
-                    serviceConfiguration.Endpoint = serviceConfiguration.Endpoint.TrimEnd('/') + "/$metadata";
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "The value \"{0}\" is not a valid MetadataDocumentUri because is it not a valid absolute Uri. The MetadataDocumentUri must be set to an absolute Uri referencing the $metadata endpoint of an OData service.", serviceConfiguration.Endpoint));
                 }
+
+                uri = CleanMetadataUri(uri);
+
+                serviceConfiguration.Endpoint = uri.AbsoluteUri;
             }
 
             Stream metadataStream;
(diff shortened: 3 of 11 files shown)


ElizabethOkerio:
@Nthemba resolve conflicts.

azure-pipelines.yml
```diff
@@ -101,10 +101,10 @@ stages:
         publishLocation: 'Container'
 
     - task: PublishBuildArtifacts@1
-      displayName: 'Publish Artifact - VSIX2022plus'
```
habbes, line 104: Why are these changes necessary?

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -185,16 +186,15 @@ private async Task<int> HandleGenerateCommand(GenerateOptions options, IConsole
             catch (Exception ex)
             {
                 console.Error.Write(ex.Message);
-                return 0;
+                return 1;
             }
 
-            return 1;
```
habbes, line 191: I thought this was already changed in a different PR?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -24,36 +26,47 @@ public static class MetadataReader
...
             Stream metadataStream;
-            Uri metadataUri = new Uri(serviceConfiguration.Endpoint);
+            var metadataUri = new Uri(serviceConfiguration.Endpoint);
 
             if (!metadataUri.IsFile)
             {
                 var webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if (serviceConfiguration.CustomHttpHeaders != null)
+                {
+                    var headerElements = serviceConfiguration.CustomHttpHeaders.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
```
habbes, line 58: Not sure I understand how this relates to the issue the PR is fixing.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +92,76 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+        public static Uri CleanMetadataUri(this Uri uri)
+        {
+            if (uri.Scheme == "http" || uri.Scheme == "https")
+            {
+                UriBuilder uriBuilder;
+
+                /// Evaluates to true if Query and Fragment properties are present in the Uri 
+                bool preserveQueryAndFragment = true;
+
+                if (uri.Segments.Last().StartsWith("$metadata", StringComparison.InvariantCultureIgnoreCase))
```
habbes, line 143: Why using `StartsWith` instead of checking that the strings match exactly? If you use `StartsWith`, that means a segment like `$metadataFoo` will be considered equivalent to `$metadata`. Is that the intent?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +92,76 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                    var absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/') + "/").Uri;
+                    uriBuilder = new UriBuilder(new Uri(absolutePathUri, "$metadata"));
+                }
+
+                if (preserveQueryAndFragment)
+                {
+                    uriBuilder.Query = uri.Query.TrimStart('?');
+                    uriBuilder.Fragment = uri.Fragment.TrimStart('#');
+                }
+                uriBuilder.UserName = uri.UserInfo;
```
habbes, line 160: nit: empty line after `}`.

```suggestion

                uriBuilder.UserName = uri.UserInfo;
```

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -54,12 +54,19 @@ public void OnPageLeavingConfigODataEndpointPageTest()
...
+            Assert.AreEqual(configOdataEndPointViewModel.UserSettings.Endpoint, "http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0#fragment");
+
             //Provide a url without $metadata
-            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService";
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://user:password@mysite/ODataService?$schemaversion=2.0#fragment";
             pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
 
-            //Check if $metadata is apended if the url does not have it added at the end
-            Assert.AreEqual(configOdataEndPointViewModel.UserSettings.Endpoint, "http://mysite/ODataService/$metadata");
+            //Check if $metadata is apended as the last segment if it was not the last segment of the url
```
habbes, line 68:
```suggestion
            //Check if $metadata is appended as the last segment if it was not the last segment of the url
```

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -54,12 +54,19 @@ public void OnPageLeavingConfigODataEndpointPageTest()
...
+            //Provide a url with "$metadata" as the last segment in the url
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0#fragment";
+            _ = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            //Check if the url is detected as valid and is unmodified
+            Assert.AreEqual(configOdataEndPointViewModel.UserSettings.Endpoint, "http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0#fragment");
+
             //Provide a url without $metadata
-            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService";
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://user:password@mysite/ODataService?$schemaversion=2.0#fragment";
```
habbes, line 65: I think it would have been better to add new test cases rather than modifying the existing ones, that way we can be confident that current scenarios are still supported.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: After a Rebase, changes made in the master branch conflicted with changes in the PR, ServiceConfiguration is used instead of UserSettings

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: The outcome is still the same i.e. a 404 ... Whether it's a 'StartsWith' or 'Equals'. the code will throw a 404 not found because the URL is not correct. Unless code is implemented to fix segments like '$metadataFoo".

azure-pipelines.yml
```diff
@@ -101,10 +101,10 @@ stages:
...
+      displayName: 'Publish Artifact - VSIX2022Plus'
```
KenitoInc, line 104: You should have rebased your branch to make this PR cleaner.
No problem though

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -24,36 +26,47 @@ public static class MetadataReader
...
 
-            
+                uri = CleanMetadataUri(uri);
+
+                serviceConfiguration.Endpoint = uri.AbsoluteUri;
+            }
 
             Stream metadataStream;
-            Uri metadataUri = new Uri(serviceConfiguration.Endpoint);
+            var metadataUri = new Uri(serviceConfiguration.Endpoint);
```
KenitoInc, line 51: why replace `Uri` with `var`?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -24,36 +26,47 @@ public static class MetadataReader
...
 
             if (!metadataUri.IsFile)
             {
                 var webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if (serviceConfiguration.CustomHttpHeaders != null)
```
KenitoInc, line 56:
```suggestion

                if (serviceConfiguration.CustomHttpHeaders != null)
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +92,77 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
             }
 
+            var workFile = Path.GetTempFileName();
+
             try
             {
                 using (XmlReader reader = XmlReader.Create(metadataStream))
                 {
-                    while (reader.NodeType != XmlNodeType.Element)
+                    using (var writer = XmlWriter.Create(workFile))
```
KenitoInc, line 101: I don't understand this change.
Why are we calling `writer.WriteNode(reader, false);`?

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -55,11 +55,32 @@ public void OnPageLeavingConfigODataEndpointPageTest()
             Assert.IsTrue(pageNavigationResult.ShowMessageBoxOnFailure);
 
             //Provide a url without $metadata
-            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService";
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService/$metadata";
```
habbes, line 58: This change contradicts the comment above it. From the above comment, I assume that this test case verified that you can pass a url without $metadata. Maybe you should create a new test case to verify that it works with $metadata, instead of modifying the existing one (unless we already have a unit test verifying that).

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -55,11 +55,32 @@ public void OnPageLeavingConfigODataEndpointPageTest()
...
+            Assert.AreEqual("http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0#fragment", configOdataEndPointViewModel.ServiceConfiguration.Endpoint);
+
+            //Provide a url with query and fragment segments without $metadata
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://user:password@mysite/ODataService?$schemaversion=2.0#fragment";
+            pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            //Check if $metadata is appended as the last segment if it was not the last segment of the url
+            Assert.AreEqual("http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0#fragment", configOdataEndPointViewModel.ServiceConfiguration.Endpoint);
+
+            //Provide a url with a fragment segment without $metadata
```
habbes, line 78: I think you could also add a test case that provides a url with $metadata and #fragment without a query string.

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -55,11 +55,32 @@ public void OnPageLeavingConfigODataEndpointPageTest()
...
-            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService";
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://mysite/ODataService/$metadata";
             pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
 
-            //Check if $metadata is apended if the url does not have it added at the end
-            Assert.AreEqual(configOdataEndPointViewModel.UserSettings.Endpoint, "http://mysite/ODataService/$metadata");
+            //Check if $metadata is appended as the last segment if it was not the last segment of the url
+            Assert.AreEqual("http://mysite/ODataService/$metadata", configOdataEndPointViewModel.ServiceConfiguration.Endpoint);
+
+            //Provide a url with "$metadata" as the last segment in the url
```
habbes, line 64:
```suggestion
            //Provide a url  with a query string and "$metadata" as the last segment in the url path
```

Nthemba:
Find updates to this PR on the link below
https://github.com/OData/ODataConnectedService/pull/311/

PR #311, opened 2022-08-19
Improve metadata uri handling

Nthemba:
This PR fixes the issue https://github.com/OData/ODataConnectedService/issues/48
and is an add on to a contributor's (https://github.com/dr-consit) PR https://github.com/OData/ODataConnectedService/pull/237

It ensures the OData Endpoint URI does not get distorted when the URI includes query and fragment segments.

Before fix
A URI like 'http://localhost:5000/odata/$metadata?api-version=1.0' would have $metadata appended to the end of the URI which would make the OData service inaccessible.

After fix
$metadata, query and fragment segments maintain the correct position and order.

### DIFF
--- a/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
+++ b/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
@@ -7,7 +7,9 @@
 
 
 using System;
+using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Xml;
 using Microsoft.OData.CodeGen.Models;
@@ -23,25 +25,27 @@ namespace Microsoft.OData.CodeGen.Common
         /// Reads the metadata version from the metadata url prvided.
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
-        /// <returns>The <see cref="Version"/> of the metadata</returns>
-        public static Version GetMetadataVersion(ServiceConfiguration serviceConfiguration)
+        /// <param name="edmxVersion">of the metadata</param>
+        /// <returns>The <see cref="String"/> of the metadata</returns>
+        public static string GetMetadataVersion(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
         {
             if (string.IsNullOrEmpty(serviceConfiguration.Endpoint))
-                throw new ArgumentNullException("OData Service Endpoint", "Input the metadata document resource");
-
-            if (File.Exists(serviceConfiguration.Endpoint))
-                serviceConfiguration.Endpoint = new FileInfo(serviceConfiguration.Endpoint).FullName;
+            {
+                throw new ArgumentNullException("OData Service Endpoint", string.Format(CultureInfo.InvariantCulture, Constants.InputServiceEndpointMsg));
+            }
 
             if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!serviceConfiguration.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out var uri))
                 {
-                    serviceConfiguration.Endpoint = serviceConfiguration.Endpoint.TrimEnd('/') + "/$metadata";
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "The value \"{0}\" is not a valid MetadataDocumentUri because is it not a valid absolute Uri. The MetadataDocumentUri must be set to an absolute Uri referencing the $metadata endpoint of an OData service.", serviceConfiguration.Endpoint));
                 }
-            }
 
-            
+                uri = CleanMetadataUri(uri);
+
+                serviceConfiguration.Endpoint = uri.AbsoluteUri;
+            }
 
             Stream metadataStream;
             Uri metadataUri = new Uri(serviceConfiguration.Endpoint);
@@ -49,6 +53,18 @@ namespace Microsoft.OData.CodeGen.Common
             if (!metadataUri.IsFile)
             {
                 var webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+
+                if (serviceConfiguration.CustomHttpHeaders != null)
+                {
+                    var headerElements = serviceConfiguration.CustomHttpHeaders.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+                    
+                    foreach (var headerElement in headerElements)
+                    {
+                        // Trim header for empty spaces
+                        var header = headerElement.Trim();
+                        webRequest.Headers.Add(header);
+                    }
+                }
 
                 if (serviceConfiguration.IncludeWebProxy)
                 {
@@ -79,32 +95,80 @@ namespace Microsoft.OData.CodeGen.Common
                 metadataStream = (Stream)xmlUrlResolver.GetEntity(metadataUri, null, typeof(Stream));
             }
 
+            var workFile = Path.GetTempFileName();
+
             try
             {
                 using (XmlReader reader = XmlReader.Create(metadataStream))
                 {
-                    while (reader.NodeType != XmlNodeType.Element)
+                    using (var writer = XmlWriter.Create(workFile))
                     {
-                        reader.Read();
-                    }
+                        while (reader.NodeType != XmlNodeType.Element)
+                        {
+                            reader.Read();
+                        }
 
-                    if (reader.EOF)
-                    {
-                        throw new InvalidOperationException("The metadata is an empty file");
-                    }
+                        if (reader.EOF)
+                        {
+                            throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "The metadata is an empty file"));
+                        }
 
-                    Constants.SupportedEdmxNamespaces.TryGetValue(reader.NamespaceURI, out var edmxVersion);
-                    return edmxVersion;
+                        Constants.SupportedEdmxNamespaces.TryGetValue(reader.NamespaceURI, out edmxVersion);
+                        writer.WriteNode(reader, false);
+                    }
                 }
+
+                return workFile;
             }
             catch (WebException e)
             {
-                throw new InvalidOperationException(string.Format("The metadata cannot be accessed"), e);
+                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, "Cannot access {0}", serviceConfiguration.Endpoint), e);
+
             }
             finally
             { 
                 metadataStream?.Dispose();
             }
+        }
+
+        /// <summary>
+        /// Ensures that the Uri conforms to the expected "$metadata" format
+        /// </summary>
+        /// <param name="uri">Uri to clean</param>
+        /// <returns>Cleaned uri</returns>
+        public static Uri CleanMetadataUri(this Uri uri)
+        {
+            if (uri.Scheme == "http" || uri.Scheme == "https")
+            {
+                UriBuilder uriBuilder;
+
+                /// Evaluates to true if Query and Fragment properties are present in the Uri 
+                bool preserveQueryAndFragment = true;
+
+                if (uri.Segments.Last().Equals("$metadata", StringComparison.InvariantCultureIgnoreCase) | uri.Segments.Last().Equals("$metadata/", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    preserveQueryAndFragment = !uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal);
+                    Uri absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/')).Uri;
+                    uriBuilder = new UriBuilder(absolutePathUri);
+                }
+                else
+                {
+                    var absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/') + "/").Uri;
+                    uriBuilder = new UriBuilder(new Uri(absolutePathUri, "$metadata"));
+                }
+
+                if (preserveQueryAndFragment)
+                {
+                    uriBuilder.Query = uri.Query.TrimStart('?');
+                    uriBuilder.Fragment = uri.Fragment.TrimStart('#');
+                }
+
+                uriBuilder.UserName = uri.UserInfo;
+
+                return new Uri(uriBuilder.Uri.AbsoluteUri);
+            }
+
+            return new Uri(uri.AbsoluteUri);
         }
     }
 }
--- a/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
+++ b/src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
@@ -22,12 +22,12 @@ namespace Microsoft.OData.CodeGen.Common
     public static class MetadataReader
     {
         /// <summary>
-        /// Reads the metadata version from the metadata url prvided.
+        /// Reads the metadata from a provided endpoint, writes to a temporary file and outputs the file path and metadata version
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
-        /// <param name="edmxVersion">of the metadata</param>
-        /// <returns>The <see cref="String"/> of the metadata</returns>
-        public static string GetMetadataVersion(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
+        /// <param name="edmxVersion">Edmx version of the metadata.</param>
+        /// <returns>Location of the metadata file.</returns>
+        public static string ProcessServiceMetadata(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
         {
             if (string.IsNullOrEmpty(serviceConfiguration.Endpoint))
             {
@@ -37,7 +37,7 @@ namespace Microsoft.OData.CodeGen.Common
             if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out var uri))
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out Uri uri))
                 {
                     throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "The value \"{0}\" is not a valid MetadataDocumentUri because is it not a valid absolute Uri. The MetadataDocumentUri must be set to an absolute Uri referencing the $metadata endpoint of an OData service.", serviceConfiguration.Endpoint));
                 }
@@ -142,20 +142,20 @@ namespace Microsoft.OData.CodeGen.Common
             {
                 UriBuilder uriBuilder;
 
-                /// Evaluates to true if Query and Fragment properties are present in the Uri 
+                // Evaluates to true if Query and Fragment properties are present in the Uri 
                 bool preserveQueryAndFragment = true;
 
                 if (uri.Segments.Last().Equals("$metadata", StringComparison.InvariantCultureIgnoreCase) | uri.Segments.Last().Equals("$metadata/", StringComparison.InvariantCultureIgnoreCase))
                 {
                     preserveQueryAndFragment = !uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal);
-                    Uri absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/')).Uri;
-                    uriBuilder = new UriBuilder(absolutePathUri);
+                    uriBuilder = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/'));
                 }
-                else
+                else if (!uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal))
                 {
-                    var absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/') + "/").Uri;
-                    uriBuilder = new UriBuilder(new Uri(absolutePathUri, "$metadata"));
+                    uriBuilder = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath + "/$metadata");
                 }
+                else uriBuilder = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath + "$metadata");
+
 
                 if (preserveQueryAndFragment)
                 {
@@ -165,10 +165,10 @@ namespace Microsoft.OData.CodeGen.Common
 
                 uriBuilder.UserName = uri.UserInfo;
 
-                return new Uri(uriBuilder.Uri.AbsoluteUri);
+                return uriBuilder.Uri;
             }
 
-            return new Uri(uri.AbsoluteUri);
+            return uri;
         }
     }
 }
--- a/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
+++ b/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
@@ -302,7 +302,7 @@ public string MetadataDocumentUri
             throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "The value \"{0}\" is not a valid MetadataDocumentUri because is it not a valid absolute Uri. The MetadataDocumentUri must be set to an absolute Uri referencing the $metadata endpoint of an OData service.", value));
         }
 
-        value = uri.CleanMetadataUri().ToString();
+        value = uri.CleanMetadataUri().AbsoluteUri;
         this.metadataDocumentUri = value;
     }
 }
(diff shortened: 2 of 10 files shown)


src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +93,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
                 metadataStream?.Dispose();
             }
         }
+
+        /// <summary>
+        /// Ensures that the Uri conforms to the expected "$metadata" format
+        /// </summary>
+        /// <param name="uri">Uri to clean</param>
+        /// <returns>Cleaned uri</returns>
+        public static Uri CleanMetadataUri(this Uri uri)
```
ElizabethOkerio, line 137: I think this method should be in the .tt file then it gets generated in the .cs file. Am not sure whether am missing something here.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -27,6 +27,7 @@ namespace Microsoft.OData.CodeGen.Templates
     using Microsoft.OData.CodeGen.Logging;
     using System.Text;
     using System.Net;
+    using Microsoft.OData.CodeGen.Common;
```
KenitoInc, line 30: Arrange namespaces in this cs file.
Begin with `System.*` then `Microsoft.*`

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -50,6 +53,17 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
             {
                 var webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
 
+                if (serviceConfiguration.CustomHttpHeaders != null)
+                {
+                    var headerElements = serviceConfiguration.CustomHttpHeaders.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
```
KenitoInc, line 58:
```suggestion
                    var headerElements = serviceConfiguration.CustomHttpHeaders.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
                    
```

test/ODataConnectedService.Tests/ViewModels/ConfigOdataEndPointViewModelTests.cs
```diff
@@ -43,23 +43,65 @@ public void OnPageLeavingConfigODataEndpointPageTest()
...
+
+            //Check if $metadata is appended as the last segment
+            Assert.AreEqual("http://user:password@mysite/ODataService/$metadata#fragment", configOdataEndPointViewModel.ServiceConfiguration.Endpoint);
+
+            //Provide a url with $metadata and a query segment without a fragment segment
+            configOdataEndPointViewModel.UserSettings.Endpoint = "http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0";
+            pageNavigationResultTask = configOdataEndPointViewModel.OnPageLeavingAsync(null);
+
+            //Check if $metadata is appended as the last segment
+            Assert.AreEqual("http://user:password@mysite/ODataService/$metadata?$schemaversion=2.0", configOdataEndPointViewModel.ServiceConfiguration.Endpoint);
```
KenitoInc, line 104: These tests are comprehensive. Good stuff!

KenitoInc (commented):
Minor changes requested.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -24,24 +26,25 @@ public static class MetadataReader
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
         /// <returns>The <see cref="Version"/> of the metadata</returns>
```
habbes, line 28: You should update the doc comments since you changed the signature of the method.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -24,24 +26,25 @@ public static class MetadataReader
...
-        public static Version GetMetadataVersion(ServiceConfiguration serviceConfiguration)
+        public static string GetMetadataVersion(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
```
habbes, line 29: What string is it returning, and why?

KenitoInc (approved):

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: The string returns a temp file path to a file that has the response from the metadata Uri given from a given connected service instance.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: This method being an extension method cannot be included in the .tt file, because it needs to be in a static class. However, we can have redesign discussion to have it accessible from other classes outside the project.

ElizabethOkerio (approved):

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -23,25 +25,27 @@ public static class MetadataReader
         /// Reads the metadata version from the metadata url prvided.
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
-        /// <returns>The <see cref="Version"/> of the metadata</returns>
-        public static Version GetMetadataVersion(ServiceConfiguration serviceConfiguration)
+        /// <param name="edmxVersion">of the metadata</param>
+        /// <returns>The <see cref="String"/> of the metadata</returns>
```
gathogojr, line 29:
```suggestion
        /// <returns>Location of the metadata file.</returns>
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -23,25 +25,27 @@ public static class MetadataReader
         /// Reads the metadata version from the metadata url prvided.
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
-        /// <returns>The <see cref="Version"/> of the metadata</returns>
-        public static Version GetMetadataVersion(ServiceConfiguration serviceConfiguration)
+        /// <param name="edmxVersion">of the metadata</param>
```
gathogojr, line 28:
```suggestion
        /// <param name="edmxVersion">Edmx version of the metadata.</param>
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -23,25 +25,27 @@ public static class MetadataReader
         /// Reads the metadata version from the metadata url prvided.
```
gathogojr, line 25:
```suggestion
        /// Reads the metadata version from the metadata URL provided.
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -23,25 +25,27 @@ public static class MetadataReader
...
         /// </summary>
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
-        /// <returns>The <see cref="Version"/> of the metadata</returns>
-        public static Version GetMetadataVersion(ServiceConfiguration serviceConfiguration)
+        /// <param name="edmxVersion">of the metadata</param>
+        /// <returns>The <see cref="String"/> of the metadata</returns>
+        public static string GetMetadataVersion(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
```
gathogojr, line 30: This method is named `GetMetadataVersion` but it returns the location of the temp metadata file instead, while the version is an out parameter. I find the naming misleading. The doc for the method is misleading too => _Reads the metadata version from the metadata url prvided._
What the method appears to be doing is read the metadata at the endpoint provided, create a temp file and save the metadata to the file, and also output the Edmx version in the process.
We should rename the method as appropriate, maybe `Parse{Service}Metadata` or `Process{Service}Metadata` or any other precise name with the location of the temp metadata file and Edmx version as outputs. The doc for the method also needs to be updated to reflect what is happening within the method.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -23,25 +25,27 @@ public static class MetadataReader
...
-                serviceConfiguration.Endpoint = new FileInfo(serviceConfiguration.Endpoint).FullName;
+            {
+                throw new ArgumentNullException("OData Service Endpoint", string.Format(CultureInfo.InvariantCulture, Constants.InputServiceEndpointMsg));
+            }
 
             if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!serviceConfiguration.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out var uri))
```
gathogojr, line 40: Please do a test with using the current version to check if we used to require people on include the `$metadata` part. If we used a user to supply just `http://localhost:80` for example and auto-fill the "$metadata` part, changing that behaviour is a breaking change

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +95,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
         }
+
+        /// <summary>
+        /// Ensures that the Uri conforms to the expected "$metadata" format
+        /// </summary>
+        /// <param name="uri">Uri to clean</param>
+        /// <returns>Cleaned uri</returns>
+        public static Uri CleanMetadataUri(this Uri uri)
+        {
+            if (uri.Scheme == "http" || uri.Scheme == "https")
```
gathogojr, line 141: Is this check necessary? What other values can we have in scheme apart from `http` or `https`? Say `uri.Scheme` is an empty string or null, should we not do the cleaning on such a uri as well?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +95,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                    uriBuilder.Query = uri.Query.TrimStart('?');
+                    uriBuilder.Fragment = uri.Fragment.TrimStart('#');
+                }
+
+                uriBuilder.UserName = uri.UserInfo;
+
+                return new Uri(uriBuilder.Uri.AbsoluteUri);
+            }
+
+            return new Uri(uri.AbsoluteUri);
```
gathogojr, line 171: Why can't we just return `uri`? What if the `uri` wasn't an absolute Uri?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +95,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+
+                if (preserveQueryAndFragment)
+                {
+                    uriBuilder.Query = uri.Query.TrimStart('?');
+                    uriBuilder.Fragment = uri.Fragment.TrimStart('#');
+                }
+
+                uriBuilder.UserName = uri.UserInfo;
+
+                return new Uri(uriBuilder.Uri.AbsoluteUri);
```
gathogojr, line 168: Why can't we just return `uriBuilder.Uri`? It'd already be an absolute Uri. I don't think there's a reason to create a new instance of `Uri`.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +95,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                if (uri.Segments.Last().Equals("$metadata", StringComparison.InvariantCultureIgnoreCase) | uri.Segments.Last().Equals("$metadata/", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    preserveQueryAndFragment = !uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal);
+                    Uri absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/')).Uri;
+                    uriBuilder = new UriBuilder(absolutePathUri);
+                }
+                else
+                {
+                    var absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/') + "/").Uri;
+                    uriBuilder = new UriBuilder(new Uri(absolutePathUri, "$metadata"));
```
gathogojr, line 157:
```suggestion
                    uriBuilder = new UriBuilder(new Uri(new Uri(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath), "$metadata"));
```
**Note:** Whether the absolute path ends with `/` or not, the `Uri(Uri, string)` constructor will do the expected thing.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +95,80 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                UriBuilder uriBuilder;
+
+                /// Evaluates to true if Query and Fragment properties are present in the Uri 
+                bool preserveQueryAndFragment = true;
+
+                if (uri.Segments.Last().Equals("$metadata", StringComparison.InvariantCultureIgnoreCase) | uri.Segments.Last().Equals("$metadata/", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    preserveQueryAndFragment = !uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal);
+                    Uri absolutePathUri = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/')).Uri;
+                    uriBuilder = new UriBuilder(absolutePathUri);
```
gathogojr, line 152:
```suggestion
                    uriBuilder = new UriBuilder(uri.AbsoluteUri.TrimEnd('/'));
```

src/ODataConnectedService.Shared/ViewModels/ConfigODataEndpointViewModel.cs
```diff
@@ -48,8 +50,10 @@ public override Task<PageNavigationResult> OnPageLeavingAsync(WizardLeavingArgs
         {
             try
             {
-                this.MetadataTempPath = GetMetadata(out var version);
+                ServiceConfiguration = GetServiceConfiguration();
+                this.MetadataTempPath = CodeGen.Common.MetadataReader.GetMetadataVersion(ServiceConfiguration, out var version);
                 // Makes sense to add MRU endpoint at this point since GetMetadata manipulates UserSettings.Endpoint
+                UserSettings.Endpoint = ServiceConfiguration.Endpoint;
```
gathogojr, line 56: Looking at how `GetServiceConfiguration` method is implemented, the values are retrieved from `UserSettings` property. What purpose does this line serve?

src/ODataConnectedService.Shared/ViewModels/ConfigODataEndpointViewModel.cs
```diff
@@ -48,8 +50,10 @@ public override Task<PageNavigationResult> OnPageLeavingAsync(WizardLeavingArgs
         {
             try
             {
-                this.MetadataTempPath = GetMetadata(out var version);
+                ServiceConfiguration = GetServiceConfiguration();
+                this.MetadataTempPath = CodeGen.Common.MetadataReader.GetMetadataVersion(ServiceConfiguration, out var version);
```
gathogojr, line 54: Would it have been better to pass `this.UserSettings` to `GetMetadataVersion` method instead of instantiating a `ServiceConfiguration` object with properties from `this.UserSettings`?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
habbes: It also does not appear that the `workFile` is used at any point in this method, nor does it use any of the method arguments. So, it appears that the creation of the temp file can safely be removed from this method altogether and that concern can be handled elsewhere, perhaps where it's needed.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: The Logic that maintains the current behavior is put in a separate method, `CleanMetadataUri`. From the tests, the behavior doesn't change for the example given above.

src/ODataConnectedService.Shared/ViewModels/ConfigODataEndpointViewModel.cs
Nthemba: `GetMetadataVersion` is accessed by both the CLI and OCS too and the CLi has Serviceconfiguration only.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: The test mentioned above passes with the current changes. The logic that maintains the behavior was moved to another method `CleanMetadataUri`

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: The file scheme is also supported. I will remove restrictions to any other schemes introduced in the future

src/ODataConnectedService.Shared/ViewModels/ConfigODataEndpointViewModel.cs
Nthemba: `UserSettings.Endpoint` is set to the updated value of `ServiceConfiguration.Endpoint` because the endpoint might be updated in the `GetServiceConfiguration` method.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -20,28 +22,31 @@ namespace Microsoft.OData.CodeGen.Common
...
-                serviceConfiguration.Endpoint = new FileInfo(serviceConfiguration.Endpoint).FullName;
+            {
+                throw new ArgumentNullException("OData Service Endpoint", string.Format(CultureInfo.InvariantCulture, Constants.InputServiceEndpointMsg));
+            }
 
             if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
                 || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
             {
-                if (!serviceConfiguration.Endpoint.EndsWith("$metadata", StringComparison.Ordinal))
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out var uri))
```
gathogojr, line 41:
```suggestion
                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out Uri uri))
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -20,28 +22,31 @@ namespace Microsoft.OData.CodeGen.Common
...
                 {
-                    serviceConfiguration.Endpoint = serviceConfiguration.Endpoint.TrimEnd('/') + "/$metadata";
+                    throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, "The value \"{0}\" is not a valid MetadataDocumentUri because is it not a valid absolute Uri. The MetadataDocumentUri must be set to an absolute Uri referencing the $metadata endpoint of an OData service.", serviceConfiguration.Endpoint));
                 }
-            }
 
-            
+                uri = CleanMetadataUri(uri);
+
+                serviceConfiguration.Endpoint = uri.AbsoluteUri;
```
gathogojr, line 48: From the explanation provided elsewhere, the argument passed into `CleanMetadataUri` may not have a `http` or `https` scheme. That would mean the returned value may not be an absolute uri - if the scheme was `file` for example. In such a case, an exception would be thrown here.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +96,79 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                bool preserveQueryAndFragment = true;
+
+                if (uri.Segments.Last().Equals("$metadata", StringComparison.InvariantCultureIgnoreCase) | uri.Segments.Last().Equals("$metadata/", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    preserveQueryAndFragment = !uri.AbsolutePath.EndsWith("/", StringComparison.Ordinal);
+                    uriBuilder = new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/'));
+                }
+                else
+                {
+
```
gathogojr, line 156:
```suggestion
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +96,79 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+                    uriBuilder = new UriBuilder(new Uri(new UriBuilder(uri.Scheme, uri.Host, uri.Port, uri.AbsolutePath.TrimEnd('/') + "/").Uri, "$metadata"));
```
gathogojr, line 157: Trimming and concatenating with `/` will result into at least 2 allocations since strings are immutable. Like I indicated in my [previous review comment](https://github.com/OData/ODataConnectedService/pull/311#discussion_r968086071), what you're doing here can be achieved more easily and with reduced allocations

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -79,32 +96,79 @@ public static Version GetMetadataVersion(ServiceConfiguration serviceConfigurati
...
+        /// </summary>
+        /// <param name="uri">Uri to clean</param>
+        /// <returns>Cleaned uri</returns>
+        public static Uri CleanMetadataUri(this Uri uri)
+        {
+            if (uri.Scheme == "http" || uri.Scheme == "https")
+            {
+                UriBuilder uriBuilder;
+
+                /// Evaluates to true if Query and Fragment properties are present in the Uri
```
gathogojr, line 146:
```suggestion
                // Evaluates to true if Query and Fragment properties are present in the Uri 
```

src/ODataConnectedService_VS2022Plus/ODataConnectedService_VS2022Plus.csproj
```diff
@@ -88,6 +88,13 @@
     <PackageReference Include="Microsoft.OData.Edm">
       <Version>7.6.3</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.SDK" Version="17.2.32505.173" ExcludeAssets="runtime">
```
gathogojr, line 91: Are the changes in this file deliberate or inadvertent?

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
gathogojr: Why can't we do like we're doing in the `.ttinclude` file
```csharp
uri = CleanMetadataUri(uri).ToString();
```
**Note:** Verify that the above would give you the equivalent of `CleanMetadataUri(uri).AbsoluteUri` where the returned value is an absolute uri.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -20,28 +22,31 @@ namespace Microsoft.OData.CodeGen.Common
     public static class MetadataReader
     {
         /// <summary>
-        /// Reads the metadata version from the metadata url prvided.
+        /// Reads the metadata from a provided endpoint and writes to a temporary file.
+        /// Reads and outputs the metadata version from the metadata URL provided.
```
gathogojr, line 26: Adopt either of these two statements or merge both to come up with a single statement

src/ODataConnectedService.Shared/ViewModels/ConfigODataEndpointViewModel.cs
Nthemba: I meant the `ProcessServiceMetadata`

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: `Uri(uri, string)` replaces the last segment if it doesn't end with `/` . For example 
`var baseUri = new Uri("http://localhost/odata");`
`var uri1 = new Uri(baseUri, "$metadata");`
will result to `http://localhost/$metadata`
That is why `/` is added before we add `"$metadata"`

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
Nthemba: There is a condition that is evaluated for each scheme with different logics

Nthemba: > Why can't we do like we're doing in the `.ttinclude` file
> 
> ```cs
> uri = CleanMetadataUri(uri).ToString();
> ```
> 
> **Note:** Verify that the above would give you the equivalent of `CleanMetadataUri(uri).AbsoluteUri` where the returned value is an absolute uri.

The value returned is the same. However, I have changed it to Absolute Uri

gathogojr: @Nthemba I actually had to confirm that `new Uri("file:///C:/DirectoryServices.csdl").AbsoluteUri` for example doesn't throw an exception. That's what I had in mind but it seems to execute just fine

gathogojr (approved):

jonaskaltenbachz:
Hello,
can you estimate when there will be a new odata-cli release for this improvements

Issue #331, opened 2023-01-04
Class property Context appears to be defined using the incorrect type of string vs DataServiceContext


wayneyan:
When generating client proxies the variable **__Context_** as defined in the C# proxy for a entity can result in compile time errors. 
To reproduce the problem, the OData metadata being used is that for Microsoft Dynamics HR. In this case all types and all bound actions are selected for generation. The following is the URL I was using with the tenant id being that relevant for my scenario. A custom heder was used to specify the bearer token for authentication purposes.

eg. https://aos-rts-sf-c8d2e0b7ff3-prod-westeurope.hr.talent.dynamics.com/namespaces/_tenant id_/data

In particular the following two compile time errors are created:

CS1061: 'string' does not contain a definition for 'EntityTracker' and no accessible extension method 'EntityTracker' accepting a first argument of type 'string' could be found (are you missing a using directive or an assembly reference?)
CS1503: Argument 1: cannot convert from 'string' to 'Microsoft.OData.Client.DataServiceContext'

and these occur with the entity type  <EntityType Name="EssWorkflowWorkItem">
where the bound action "complete" results in generation of the code 

```
        [global::Microsoft.OData.Client.OriginalNameAttribute("complete")]
        public virtual global::Microsoft.OData.Client.DataServiceActionQuery Complete(string outcome, string comment)
        {
            global::Microsoft.OData.Client.EntityDescriptor resource = Context.EntityTracker.TryGetEntityDescriptor(this);
            if (resource == null)
            {
                throw new global::System.Exception("cannot find entity");
            }

            return new global::Microsoft.OData.Client.DataServiceActionQuery(this.Context, resource.EditLink.OriginalString.Trim('/') + "/Microsoft.Dynamics.DataEntities.complete", new global::Microsoft.OData.Client.BodyOperationParameter("outcome", outcome),
                    new global::Microsoft.OData.Client.BodyOperationParameter("comment", comment));
        }
```

On inspection the property **_Context_** appears to be type string but should be type **_DataServiceContext_**. The following snippet is the generated fragment.

```
        [global::Microsoft.OData.Client.OriginalNameAttribute("Context")]
        public virtual string Context
        {
            get
            {
                return this._Context;
            }
            set
            {
                this.OnContextChanging(value);
                this._Context = value;
                this.OnContextChanged();
                this.OnPropertyChanged("Context");
            }
        }

```
There is also the definition of a property **_ContextString_** which I believe to be causing some confusion in the template.

```
[...]

Issue #317, opened 2022-09-08
Dynamics 365 OData metadata contains spaces causing OData generated code to fail compilation


pmkevans:
The issue is described here: https://community.dynamics.com/365/financeandoperations/f/dynamics-365-for-finance-and-operations-forum/463165/odata-properties-have-spaces
In short, current versions of D365 have spaces in some property names.  OData-CLI is generating class and variable names using this information without managing the spaces and thereby generating code that does not compile.

Although the root cause is D365 being noncompliant it would be sensible for ODataConnectedService to have smarter name generation to validate that all names generated are valid C# identifiers.

mikepizzo:
We could add logic to the class generation (i.e., to replace space with underscore) but that could be ambiguous (i.e., if the store had the same property name in one case with a space and in one case without a space).  We could add additional handling for that case, or even just consider it an edge case and "better than what we have now", however, generating the types with properties that don't contain a space is just a small step in what is going to be a much bigger issue, which is the fact that the ODataClient (or any other client using the classes) is going to have to create URLs to query the service, and read/bind responses to the types, and all of the underlying libraries for handling OData Urls and payloads aren't going to be able to deal with property names that contain spaces.

sanderson196:
As a follow up to the initial Feature request, I have (manually) modified the generated c# code to remove the spaces from the generate method names and where they are used but not removing the spaces from code related to the original Odata Endpoints.  For example, ((this._Template Tasks == null)) is converted to ((this._TemplateTasks == null)) but Context.CreateQuery<global::Microsoft.Dynamics.DataEntities.ProcessTemplateTask>(GetPath("Template tasks")); remains the same.

Once done you get compliable and usable code that appears to still be able to access the OData endpoints properly.  I encourage the community to verify my results.

For reference, I use a regex search to find and replace the "offending" c# code
(?<!")Template tasks(?!")

magiva:
has there been any progress on a fix for this ? who owns the issue here? as microsoft introduced the issue and isnt this tool also microsofts ?

magiva:
do you have any links to the process you used or did you just use regex to replace the space with _ or remove etc ?

sanderson196:
Just used Regex for search for and replace.  The catch is that you can't change the data between quotes (the reference to Dynamcis) but only change the code function references.  Therefore your regex has to exclude items that start and end in "

Scott Anderson
IT Manager


Mobile: +1 9192916265   |  Phone: 919.804.6331
[...]

Issue #297, opened 2022-06-08
odata-cli generated entity with "Context" property will not compile


benbward:
### Describe the bug
When the entity includes a property named, "Context", a conflict arises with the Microsoft.OData.Client.BaseEntityType.Context object of the same name.

#### Version of the Project affected
Microsoft.OData.Cli v0.1.0
VisualStudio Version v17.2.3
Microsoft.Odata.Client v7.11.0

#### To Reproduce
1. Define an OData api with an entity having a "Context" property.  (`<Property Name="Context" Type="Edm.String">`)
2. Use the OData CLI to generate classes.
3. Build the project.
4. Observe errors including `'string' does not contain a definition for 'EntityTracker' `
5. Observe a warning `'EssWorkflowWorkItem.Context' hides inherited member 'BaseEntityType.Context'.`

#### Expected behavior
No compilation errors should occur.

#### Actual behavior
Compilation errors occur.

#### Additional context
We are working around this by removing the generated (and unused) method where the error occurs.

Issue #242, opened 2015-07-27
[T4] OData v4 Client Code Generator generates a  C# proxy class that won’t compile if the model contains an Entity that has a property with name "Context"


LaylaLiu:
BaseEntityType contains a property named "Context".

So when a model contains an entity which has a property named "Context". The client generated code will be like 

public partial class Product : global::Microsoft.OData.Client.BaseEntityType
{
            ...
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "2.1.0")]
        [global::Microsoft.OData.Client.OriginalNameAttribute("Context")]
        public global::System.Collections.ObjectModel.ObservableCollectionglobal::OutlookApiBetaModel.CustomProperty Context
        {
            get            {                ...            }
            set            {                ...            }
        }

This will lead to "ConsoleApplication7.Product.Context' hides inherited member 'Microsoft.OData.Client.BaseEntityType.Context'. Use the new keyword if hiding was intended. “

mikepizzo:
We probably should have named these internal properties such that they had a lower chance of conflicting with user properties.  i.e., "__context" or some such.

PR #333, opened 2023-02-02
Create valid property identifiers when generating OData client

gregwinterstein:
Modify the code generator to generate valid identifiers

This change resolves the following issues:
* #317
* #297 & #331
* #242
* #224


gregwinterstein:

jackypowell:
My team really needs these changes. Can we please get them reviewed soon? Please and thank you! @ElizabethOkerio @KenitoInc @habbes @gathogojr

habbes:
@gathogojr you had some thoughts around this issue. Could you check this PR?

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2442,7 +2444,9 @@ internal void WritePropertiesForSingleType(IEnumerable<IEdmProperty> properties)
         foreach (IEdmProperty property in properties.Where(i => i.PropertyKind == EdmPropertyKind.Navigation))
         {
             string propertyType;
-            string propertyName = this.context.EnableNamingAlias ? Customization.CustomizeNaming(property.Name) : property.Name;
+            string propertyName = IdentifierMappings.ContainsKey(property.Name) ?
```
gathogojr, line 2447: Use `Dictionary<TKey,TValue>.TryGetValue(TKey, TValue)` method instead of `ContainsKey` method + `Dictionary<TKey,TValue>.Item[TKey]` property. `TryGetValue` should be more efficient and cleaner

gathogojr:
@gregwinterstein Have you considered that this would be a breaking change for someone who's REGENERATING the proxy classes?

gregwinterstein:
> Have you considered that this could be a breaking change for someone who's REGENERATING the proxy classes? Specifically for the case of the property named `Context`.
> 
@gathogojr, it has occurred to me that this pull request may cause some breaking changes because of the way valid identifiers are generated vs the way they were generated previously. I don't have a good solution for that.

My main issue is that I am unable to generate proxies that compile for either Dynamics F&O or Dynamics CRM (this doesn't entirely resolve all of the CRM issues due to inconsistent casing in the CRM metadata). This change generates proxies that compile for Dynmaics F&O and gets a lot closer for Dynamics CRM.

If backward compatibility is an issue, I'm open to suggestions.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
gregwinterstein: Changed to `TryGetValue()` in [a869504](https://github.com/OData/ODataConnectedService/pull/333/commits/a8695043b915eddf2ef58ced602c46638fe4e387)

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
```diff
@@ -2803,6 +2805,187 @@ public void WriteDupNamesWithCamelCase()
 
         #endregion
 
+        #region Tests for invalid identifiers
+
+        private const string InvalidNamesEdmx = @"<?xml version=""1.0"" encoding=""utf-8""?>
```
gathogojr, line 2810: @gregwinterstein I think you're implementing a fix for a invalid/impossible scenarios.
- [The edm:Property element MUST include a Name attribute whose value is a [SimpleIdentifier](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_SimpleIdentifier) used when referencing, serializing or deserializing the property.](https://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752526)
- [Non-normatively speaking a `SimpleIdentifier` starts with a letter or underscore, followed by at most 127 letters, underscores or digits](https://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_SimpleIdentifier)
 
For that reason, the following will never be valid property names in a CSDL document and we shouldn't try to find a way to sanitize them:
- "Name with Spaces"
- "Invalid$characters"
- "Invalid ! characters"

We should instead focus on valid scenarios that lead to compilation issues.
The best example being a property named `Context` while there's a similarly named property defined in the `IBaseEntityType` interface.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,156 @@ public abstract class ODataClientTemplate : TemplateBase
...
+    /// </summary>
+    /// <remarks>
+    /// C# and VB use very similar rules for identifier names
+    /// </remarks>
+    /// <param name="name">Identifier to validate and modify</param>
+    /// <returns>Modified <paramref name="name"/> or <paramref name="name"/> if no changes are needed</returns>
+    private string GetValidIdentifier(string name)
+    {
+        var validName = name;
+        if (!string.IsNullOrWhiteSpace(validName))
```
gathogojr, line 2900: No need for this check. This scenario should never arise

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,156 @@ public abstract class ODataClientTemplate : TemplateBase
...
+        {
+            // Spaces are not permitted, capitalize and strip spaces
+            if (validName.Contains(" "))
+            {
+                validName = CultureInfo.CurrentCulture.TextInfo.ToTitleCase(validName);
+                validName = validName.Replace(" ", "");
+            }
+
+            // Replace non-alphanumeric characters
+            validName = RegexMatchNonIdentifier.Replace(validName, DefaultReplacementCharacter);
```
gathogojr, line 2910: See my comment here https://github.com/OData/ODataConnectedService/pull/333/files#r1374567249

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,156 @@ public abstract class ODataClientTemplate : TemplateBase
...
+
+            // Prefix initial numbers with an underscore to meet requirements
+            if (RegexMatchNumericStart.IsMatch(validName))
```
gathogojr, line 2913: See my comment here https://github.com/OData/ODataConnectedService/pull/333/files#r1374567249.
A property name will never start with a number

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,156 @@ public abstract class ODataClientTemplate : TemplateBase
...
+            {
+                validName = $"{DefaultReplacementCharacter}{validName}";
+            }
+
+            // Capitalize identifier for consistency
+            if (this.context.EnableNamingAlias && validName.First() != '_' && validName.First() != '@' && !char.IsUpper(validName.First()))
```
gathogojr, line 2919: A property name will never start with an `@` symbol

gathogojr, line 2919: A simple strategy for dealing with the property naming conflict might be just concatenating the type name with the property name.
For example, for a `Context` property in a `Product` class, we can concatenate that to get `ProductContext` and apply the `OriginalName` attribute to indicate the original name is `Context`.
I think a public property starting with an `_` might be undesirable.

**Important** to note:
- Be careful about camel-cased properties so we don't end up with `productcontext` while the more desirable outcome would be `productContext` (I think).
- Be careful when thinking about case because in VB `Context` and `context` wouldn't be considered different.

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
gregwinterstein: @gathogojr I disagree. Most of these issues are found in Microsoft Products.
Both Microsoft AX (F&O) and Microsoft CRM provide metadata that does not conform to these requirements.
We cannot use the ODataConnectedService to generate proxy classes for AX & CRM without these proposed changes.
The generated code does not compile and includes many the issues noted above.

gathogojr: @gregwinterstein
> Most of these issues are found in Microsoft Products

Is it impossible for those issues to be fixed from those products/libraries?

Let's leave OData Connected Service extension aside for a minute and imagine an OData service that has properties that don't conform to the OData protocol. For instance, one that has a property name with a space - "Account Number".
If you need to query such a service from a browser or an API client like Postman, how would you write a valid request URI to say filter by that property? `$filter=Account Number eq 'foobar'` would be invalid.
Outside of OData Connected Service and OData client application, how does someone interact with such a service if it doesn't adhere to the property naming rules?
- https://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752526
- https://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_SimpleIdentifier

Another question is, why should the OData team place upon itself the responsibility of curing issues that may manifest in all manner of unexpected and uncontrolled ways? Your pull request cures only a small subset of possible naming violations. How would we keep up?

gathogojr: @gregwinterstein I believe @mikepizzo was raising the same issue in this comment https://github.com/OData/ODataConnectedService/issues/317#issuecomment-1245700905. Where property names don't adhere to the OData rules on identifiers, how are such properties applied to generate valid request OData URIs?

gathogojr: @SomeTroglodyte I know for a fact the OData URI parser will not successfully parse the request URI if a property name has a space (for example). How is asking how that is handled politics?

gregwinterstein: The implementation of OData is up to the implementer. There are non-conformist "OData implementations" in existence (including some of Microsoft's own products among them). My current use cases are MS Dynamics AX (F&O) and MS Dynamics CRM. Both of which claim to implement OData. The metadata from these applications used with the current production ODataConnectedService extension result in generated code that does not compile. I do not have time to manually adjust the generated code each time the code is regenerated (and I do not have time to maintain a fork of this extension). This is the reason for my proposed changes.

Setting all of that aside.

I have adjusted the mechanism for generating valid identifiers to follow the C# & VB identifier specifications as of this commit [fa5c1a7](https://github.com/OData/ODataConnectedService/pull/333/commits/fa5c1a7c739eb8f4447048a5fe3a5a559a248a1d).

This generates identifiers that compile without the need to keep up with all possible permutations of invalid identifiers.

Note: This change introduces a difference in behavior. The original used underscores in place of invalid characters. This update removes the underscores from the generated identifiers.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
gregwinterstein: Implemented this great suggestion in [d3b26b0](https://github.com/OData/ODataConnectedService/pull/333/commits/d3b26b091ca143075b42f7090f943802cbd83c85)

gregwinterstein: Removed.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
gregwinterstein: Removed.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,156 @@ public abstract class ODataClientTemplate : TemplateBase
...
+    /// </remarks>
+    /// <param name="name">Identifier to validate and modify</param>
+    /// <returns>Modified <paramref name="name"/> or <paramref name="name"/> if no changes are needed</returns>
+    private string GetValidIdentifier(string name)
+    {
+        var validName = name;
+        if (!string.IsNullOrWhiteSpace(validName))
+        {
+            // Spaces are not permitted, capitalize and strip spaces
+            if (validName.Contains(" "))
```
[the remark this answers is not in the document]
gregwinterstein: Removed in this commit [fa5c1a7](https://github.com/OData/ODataConnectedService/pull/333/commits/fa5c1a7c739eb8f4447048a5fe3a5a559a248a1d).

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
gathogojr: @gregwinterstein Apologies for going quiet on this. We had a discussions and did practical tests to confirm how a scenario with disallowed characters in property names would arise. We realized that dynamic properties practically make it possible and if the `OriginalName` attribute contains a property name with disallowed character, deserialization works just fine. We are convinced allowing this fix does more good than harm under the circumstances.
We need to identify a practical subset of special characters to accommodate in addition to ensuring the logic for sanitizing the property name to generate a valid identifier is efficient.

Here's code for one scenario that we tested...

A service that returns a dynamic property with spaces - "Date Of Birth" - (or invalid property names in general):
```csharp
// Service configuration
using Microsoft.AspNetCore.OData;
using Microsoft.OData.Edm;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddControllers().AddOData(
    options => options.AddRouteComponents(
        model: GetEdmModel()));

var app = builder.Build();

app.UseRouting();
app.MapControllers();

app.Run();

IEdmModel GetEdmModel()
{
    var model = new EdmModel();

    var personEntityType = new EdmEntityType("Ex282.Models", "Person", baseType: null, isAbstract: false, isOpen: true);
    personEntityType.AddKeys(personEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32, isNullable: false));
    personEntityType.AddStructuralProperty("Name", EdmPrimitiveTypeKind.String, isNullable: false);
    model.AddElement(personEntityType);

    var container = model.AddEntityContainer("Default", "Container");

    container.AddEntitySet("People", personEntityType);

    return model;
}

// Controller
using Microsoft.AspNetCore.OData.Extensions;
using Microsoft.AspNetCore.OData.Formatter.Value;
using Microsoft.AspNetCore.OData.Routing.Controllers;
using Microsoft.OData.Edm;
using Microsoft.OData.UriParser;

namespace Ex282.Controllers
{
    public class PeopleController : ODataController
    {
        public EdmEntityObjectCollection Get()
        {
            IEdmCollectionType collectionType = (IEdmCollectionType)Request.ODataFeature().Path.EdmType().Definition;
            IEdmEntityTypeReference elementType = collectionType.ElementType.AsEntity();

            // Create an untyped collection with the Edm collection type
            EdmEntityObjectCollection collection = new EdmEntityObjectCollection(new EdmCollectionTypeReference(collectionType));

            var entityObject1 = new EdmEntityObject(elementType);
            entityObject1.TrySetPropertyValue("Id", 1);
            entityObject1.TrySetPropertyValue("Name", "Sue");
            entityObject1.TrySetPropertyValue("Date Of Birth", "2000-07-13");

            var entityObject2 = new EdmEntityObject(elementType);
            entityObject2.TrySetPropertyValue("Id", 2);
            entityObject2.TrySetPropertyValue("Name", "Joe");
            entityObject2.TrySetPropertyValue("Date Of Birth", "2001-05-17");

            // Add untyped objects to the collection
            collection.Add(entityObject1);
            collection.Add(entityObject2);

            return collection;
        }
    }
}
```

Then generate service proxy using **OData Connected Service** and use the `OriginalName` attribute to map the property name with space to a valid C# identifier:
```csharp
// Partial class - to define properties to be mapped to the dynamic properties in the response
using Microsoft.OData.Client;

namespace Ex282.Models
{
    public partial class Person
    {
        [OriginalName("Date Of Birth")]
        public string? DateOfBirth { get; set; }
    }
}

// Code to retrieve people
using Default;

var serviceRoot = new Uri("http://localhost:5000");
var dataServiceContext = new Container(serviceRoot);

var people =  dataServiceContext.People.ToArray();
```

The `DateOfBirth` property was properly materialized even though the service returns it as `Date Of Birth`:
image

gathogojr: @gregwinterstein Let's use a simple substitution routine. Unless we have no other option, let's not use Regex

gregwinterstein: @gathogojr The C# (and VB) identifier specification is very liberal in what is permitted as valid identifier characters (essentially large chunks of Unicode). It is defined by regex Unicode character classes. It is for this reason that I used a regex.

[C# identifier specification](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers)
[VB identifier specification](https://learn.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/lexical-grammar#identifiers)
[Unicode character class definitions](https://www.unicode.org/reports/tr18/#General_Category_Property)

I am open to suggestions for how to identify characters that must be replaced. I would be happy to use a simple substitution. But, I don't have any other ideas on how to identify the myriad of valid (or invalid) identifier characters aside from the regex derived from the C# specification.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2908,6 +2919,166 @@ internal void SetPropertyIdentifierMappingsIfNameConflicts(string typeName, IEdm
...
+            var titleCaseSegment = Customization.CustomizeNaming(segment);
+
+            if (isFirst && !this.context.EnableNamingAlias)
+            {
+                validName.Append(segment);
+            }
+            else
+            {
+                validName.Append(titleCaseSegment);
+            }
```
habbes, line 3069: Perhaps we should only call `Customization.CustomizeNaming(segment)` when necessary
```suggestion
            if (isFirst && !this.context.EnableNamingAlias)
            {
                validName.Append(segment);
            }
            else
            {
                var titleCaseSegment = Customization.CustomizeNaming(segment);
                validName.Append(titleCaseSegment);
            }
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2908,6 +2919,166 @@ internal void SetPropertyIdentifierMappingsIfNameConflicts(string typeName, IEdm
...
+    /// See C# definitions: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers <br />
+    /// See VB definitions: https://learn.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/lexical-grammar#identifiers <br />
+    /// See Unicode character class definitions: https://www.unicode.org/reports/tr18/#General_Category_Property
+    /// </summary>
+    private static readonly Regex RegexMatchNonIdentifier = new Regex(@"[^\p{L}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\p{Cf}]", RegexOptions.Compiled);
+
+    /// <summary>
+    /// Default prefix character used to prefix non-alpha characters in identifiers
+    /// </summary>
+    private const string DefaultPrefixCharacter = "_";
```
habbes, line 3033: Since it's only a single character, is it necessary to make it a string?

```suggestion
    private const charDefaultPrefixCharacter = '_';
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -2908,6 +2919,166 @@ internal void SetPropertyIdentifierMappingsIfNameConflicts(string typeName, IEdm
...
+            }
+            else
+            {
+                validName.Append(titleCaseSegment);
+            }
+
+            isFirst = false;
+        }
+
+        if (!char.IsLetter(validName.ToString(), 0) && !validName.ToString().StartsWith(DefaultPrefixCharacter))
```
habbes, line 3074: Indexing the string builder directly would be far more efficient than doing `ToString()` just to read the first char. We're also calling `ToString()` multiple times which feels unnecessarily wasteful.

```suggestion
        if (!char.IsLetter(validName[0]) && validName[0] != DefaultPrefixCharacter)
```

image
Benchmark code: https://gist.github.com/habbes/f244a5991f56196f2c3030ecb2a7f8fb

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,166 @@ public abstract class ODataClientTemplate : TemplateBase
...
+    /// See C# definitions: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers <br />
+    /// See VB definitions: https://learn.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/lexical-grammar#identifiers <br />
+    /// See Unicode character class definitions: https://www.unicode.org/reports/tr18/#General_Category_Property
+    /// </summary>
+    private static readonly Regex RegexMatchNonIdentifier = new Regex(@"[^\p{L}\p{Nl}\p{Mn}\p{Mc}\p{Nd}\p{Pc}\p{Cf}]", RegexOptions.Compiled);
+
+    /// <summary>
+    /// Default prefix character used to prefix non-alpha characters in identifiers
+    /// </summary>
+    private const string DefaultPrefixCharacter = "_";
```
habbes, line 2888:
```suggestion
    private const char DefaultPrefixCharacter = '_';
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,166 @@ public abstract class ODataClientTemplate : TemplateBase
...
+            var titleCaseSegment = Customization.CustomizeNaming(segment);
+
+            if (isFirst && !this.context.EnableNamingAlias)
+            {
+                validName.Append(segment);
+            }
+            else
+            {
+                validName.Append(titleCaseSegment);
+            }
```
habbes, line 2924:
```suggestion
            if (isFirst && !this.context.EnableNamingAlias)
            {
                validName.Append(segment);
            }
            else
            {
                var titleCaseSegment = Customization.CustomizeNaming(segment);
                validName.Append(titleCaseSegment);
            }
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,166 @@ public abstract class ODataClientTemplate : TemplateBase
...
+
+            isFirst = false;
+        }
+
+        if (!char.IsLetter(validName.ToString(), 0) && !validName.ToString().StartsWith(DefaultPrefixCharacter))
+        {
```
habbes, line 2930:
```suggestion
        if (!char.IsLetter(validName[0]) && validName[0] != DefaultPrefixCharacter)
        {
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -2763,6 +2774,166 @@ public abstract class ODataClientTemplate : TemplateBase
...
+            }
+            else
+            {
+                validName.Append(titleCaseSegment);
+            }
+
+            isFirst = false;
+        }
+
+        if (!char.IsLetter(validName.ToString(), 0) && !validName.ToString().StartsWith(DefaultPrefixCharacter))
```
habbes, line 2929: Are we guaranteed that `validName` cannot be empty at this point?

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
habbes: Since the regex is only used during code generaton and not a runtime, and is a static and compiled regex, I'm fine with us using it.

habbes:
@gregwinterstein I'm fine with the PR in general. I've left a few comments where I think the code is unnecessarily wasteful and could be improved.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
gregwinterstein: It is unlikely that we would have an empty `name` being passed to `GetValidIdentifier(name)`. If so, it would suggest that one of the properties provided by IEdmStructuredType.DeclaredProperties was null/empty. I suppose that could happen, but it would suggest a defective source model.

If there are no null/empty `IEdmStructuredType.DeclaredProperties`, then `foreach (var segment in segments.Where(token => !string.IsNullOrWhiteSpace(token)))` will provide at least one non null/empty `segment` and the if statements inside the `foreach` will append one of those segments to `validName`.

That said, I've added a check and throw an exception if no valid identifier was produced in [205a83f](https://github.com/OData/ODataConnectedService/pull/333/commits/205a83f2e199f834b066be2af1e4f12dd6b73226)

habbes:
> > unnecessarily wasteful
> 
> Don't forget source readability trumps performance. Here in this context. By far. No matter how wasteful a toString is, this will _always_ be far less sluggish than, say, ~censored~. And overall performance is in my eyes controlled mostly by the ODATA service in question, the tool not so much. And tool runtime is only a fraction of resulting product runtime.
> 
> Of course, one of these proposals **does** read way better...

I would argue that the proposed changes, in addition to being more performant, were as readable (if not more) than the original code. So in the case where readability is similar, and the cost to create more performant code is low (didn't require a lot of refactoring), then there's no reason not to go for the more performant code.

For example, `char.IsLetter(validateName[0])` is not any less readable than `char.IsLetter(validName.ToString(), 0)`. I would argue that it's in fact more readable, it's clear that your checking whether the first character is a letter. In the second snippet (the original code), it may not be immediately clear what the second argument of the `char.IsLetter` method refers to. The code change is straightforward and introduces virtually no risks. So in this case, there's no reason not to switch to the more performant code in my opinion.

habbes (approved):

gathogojr (approved):

gathogojr:
Thank you @gregwinterstein for your contribution

Issue #269, opened 2022-04-05
odata-cli tracking option is missing


artur-moreira:
As per #247

--no-tracking |-n - disables entity and property tracking

Should assume that UseDataServiceCollection would be true, but never assigned or supported on CLI commands, so defaults to false producing a proxy without change tracking enabled

andreas-developer:
Any progress on this issue? We're using the D365 F&O OData interface and the CLI is the only option that doesn't require a machine with 40+ Gb RAM when generating the client. But without the tracking we're losing a lot of functionality in our existing integrations.

bradschoen:
We too prefer the CLI for multiple reasons and cannot use it because of this issue.

ElizabethOkerio:
@bradschoen @andreas-developer we'll check on this as soon as possible.

suomimar-low:
Will there be a fix for this? I see the PR is almost ready?

suomimar-low:
We are using OData with Business Central and for that use Odata CLI for automatic proxy class generation in CI. But we cannot use this client generator now because we cannot do any updates via data context because property tracking is disabled. All Business Central entities contain such properties that are read only. We would need to do manual property tracking which would not make sense in this context - if that would work either.

batetech:
I ran into this same issue when working on a .NET project that is using Dynamics 365 data entities which needs upgraded to use VS2019/VS2022 instead of VS2017.  Since PR #330 has been stuck for 2 months because it is out-of-date with the base branch, I went ahead and implemented this as a new PR #339 which adds a new option --use-data-service-collection (-dsc) to add the entity/property tracking.  The up-side of adding the new option this way (instead of as --no-tracking) is that it will not break or change the default for anyone that is already using odata-cli to generate their code, but will give others the option to opt in to the property tracking when needed.  Hopefully this can get reviewed and included at some point in the future. 

@ElizabethOkerio this is my first PR on a public project, so please let me know if there is anything I missed.

marcmognol:
Hi, you are right my PR should be rebased, but as I didn't get any feedback for it, it looked like this project don't accept any new PR.

Please tell me if I have to rebase it, and if this project is still alive.

Thanks

suomimar-low:
Hi! I think it might be good to rebase? Next time they have to review the PR, it would contain up-to-date code. For us this PR is a must so thanks for doing this. 👍

marcmognol:
Hi @suomimar-low I've rebased my branch.

[...]

PR #339, opened 2023-02-13
Add --use-data-service-collection (-dsc) option

batetech:
Add --use-data-service-collection (-dsc) option which aligns with the "Enable Entity and Property Tracking" advanced option in the VS Extension.  Related to issue #269 but I implemented this as a new option --use-data-service-collection instead of --no-tracking, so that it is not a change in the defaults for existing odata-cli users.  Users that would like to opt in to this functionality can use the new parameter.

### DIFF
--- a/src/Microsoft.OData.Cli/GenerateCommand.cs
+++ b/src/Microsoft.OData.Cli/GenerateCommand.cs
@@ -69,6 +69,14 @@ namespace Microsoft.OData.Cli
 
             this.AddOption(ns);
 
+            Option useDataServiceCollection = new Option<bool>(new[] { "--use-data-service-collection", "-dsc" })
+            {
+                Name = "use-data-service-collection",
+                Description = "Allows the DataServiceContext to keep track of changes made to entities on the client side. This option generates events that are triggered when properties of the entities change. Aligns with the \"Enable Entity and Property Tracking\" advanced option in the VS Extension."
+            };
+
+            this.AddOption(useDataServiceCollection);
+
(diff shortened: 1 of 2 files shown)


batetech:

ElizabethOkerio:
@batetech thanks for the contribution. So, we currently do not have `--no-tracking |-n` in the CLI command options but it is documented. I'd suggest we use this name instead of this `--use-data-service-collection` as it is already documented.

batetech:
@ElizabethOkerio the reason I didn't use --no-tracking is PowerShell switches are normally false by default and true when included, so if we introduce --no-tracking with a default of false then this change would change the resulting code for all users that currently use odata-cli and would be a potential breaking change. If we leave the tracking off by default and give the new option to opt in then this wouldn't require existing users to change anything to their current setups/scripts that use odata-cli.  

Based on this, do you still prefer the --no-tracking option instead?

batetech:
@ElizabethOkerio where is --no-tracking documented?  I couldn't find it and isn't on the docs page here https://learn.microsoft.com/en-us/odata/odatacli/getting-started  or a Google search https://www.google.com/search?q=%22odata-cli%22+%22no-tracking%22  

"UseDataServiceCollection": true is also what the ConnectedService.json file uses via the VS Extension, which is another reason I used this name.

habbes:
@batetech If by default tracking is not enabled, I think `--enable-tracking` would be a more suitable name for the command since it describes the feature better whereas `--use-data-service-collection` exposes unnecessary implementation details without making clear what the impact is. I think it's unfortunate that we `UseDataServiceCollection` in the `ConnectedService.json` file, but I don't think we have to use the same name for user-facing command flags.

PR #345, opened 2023-04-27
Fix default value initialization for nullable properties

### LINKED ISSUE #342: Wrong handling of default value of Edm.Date property
### Describe the bug
A default value of a property of type `Edm.Date` is wrongly handled as an integer.

ODataConnectedService 2022+ v1.0.1
Microsoft Visual Studio Professional 2022 (64-bit) v17.5.4
Microsoft.Odata.Client v7.15.0

#### To Reproduce
Import the following OData V4 metadata:
```
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
    <Schema Namespace="self" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityContainer Name="EntityContainer">
        <EntitySet Name="Dates" EntityType="self.Date" />
      </EntityContainer>
      <EntityType Name="Date">
        <Key>
          <PropertyRef Name="id" />
        </Key>
        <Property Name="id" Type="Edm.Guid" Nullable="false" />
        <Property Name="date" Type="Edm.Date" DefaultValue="9999-12-31" />
      </EntityType>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```

#### Expected behavior
Generation of a working C# OData client.

#### Actual behavior
The following line of C# code is produced:
```
private global::System.Nullable<global::Microsoft.OData.Edm.Date> _Date1 = 9999-12-31;
```
in which the default value for the Edm.Date property seems to be an arithmetic expression evaluating to the integer 9956. This leads to the error
> Cannot implicitly convert type 'int' to 'Microsoft.OData.Edm.Date?'

#### Screenshots
n/a

#### Additional context
n/a

habbes:
## Problem

Fix #342 

When a property definition in the CSDL document specifies a default value like follows:

```xml
<Property Name="Date" Type="Edm.Date" Nullable="false" DefaultValue="2020-02-02" />
```
The generated code (for both C# and VB) includes a corresponding initialization for the backing value of the CLR property:
```c#
private global::Microsoft.OData.Edm.Date _Date = global::Microsoft.OData.Edm.Date.Parse("2020-02-02");
```

However, when the CSDL property is nullable, the initializer is not properly generated:

```xml
<Property Name="Date" Type="Edm.Date" DefaultValue="2020-02-02" />
```

```c#
private global::System.Nullable<global::Microsoft.OData.Edm.Date> _NullableDateProp = 2020 - 02 - 02;
```
Depending on the property type and initialization, this may lead to errors. For example, the above expression will evaluate to an integer instead of a date, causing a compile-time error.

## Solution

The code generator generates different initialization code based on the data type. However, its data-type detection logic does not take into Nullable types into account. A boolean EDM property with `Nullable` set to true will be translated to `Nullable<bool>` in .NET instead of `bool`. The solution makes adjustments to this logic such that `Nullable<T>` and `T` are treated the same way for the purpose of property value initialization.


src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3723,118 +3732,119 @@ internal static string GetPropertyInitializationValue(IEdmProperty property, boo
...
+            // The property initializers below do not expect that the specified default to be null. Since the default
+            // value of a Nullable<T> is null, it's not worth the effort to handle that edge case, Instead we
+            // assume the user will not specify "null" as the default value of a nullable property since it's redundant.
+            valueClrType = clientTemplate.StripNullableFromTypeName(valueClrType);
+            string defaultValue = structuredProperty.DefaultValueString;
+            bool isCSharpTemplate = clientTemplate is ODataClientCSharpTemplate;
+            if (edmTypeReference.Definition.TypeKind == EdmTypeKind.Enum)
+            {
+                var enumValues = defaultValue.Split(',');
+                string fullenumTypeName = GetClrTypeName(edmTypeReference, useDataServiceCollection, clientTemplate, context);
```
ElizabethOkerio, line 3758:
```suggestion
                string fullEnumTypeName = GetClrTypeName(edmTypeReference, useDataServiceCollection, clientTemplate, context);
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1745,6 +1745,15 @@ public override int GetHashCode()
...
     #endregion Language specific write methods.
 
+    /// <summary>
+    /// Given a nullable type name, e.g. Nullable<T>, returns
+    /// the underlying type name T. If the type name is not
+    /// a Nullable<T>, then returns T.
+    /// </summary>
+    /// <param name="typeName">The type name.</param>
+    /// <returns>The underlying type name.</returns>
+    internal abstract string StripNullableFromTypeName(string typeName);
```
ElizabethOkerio, line 1755: so the type name here will be in form of a string? Exa `Nullable<DateTime>` then  the method returns `DateTime` but in string form?

ElizabethOkerio: Maybe I do not understand but why can't we check if a type is  nullable without the conversions to string?

habbes: Because by the time we get to use this method, the input type name is already in string form. I do agree that it would be more robust to use actual reflection type metadata to detect nullability, but we don't have that information in the method where this method is used. To use types would refactoring a lot more methods and be a lot more work than what was allocated to fix this bug.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -3723,118 +3732,119 @@ internal static string GetPropertyInitializationValue(IEdmProperty property, boo
...
-                            // decimal in VB must be initialized with 'D' at the end, like Decimal dec = 3.00D
-                            defaultValue = defaultValue.ToLower(CultureInfo.InvariantCulture).Replace("m", "D");
-                            defaultValue = defaultValue.EndsWith("D", StringComparison.OrdinalIgnoreCase) ? defaultValue : defaultValue + "D";
-                        }
-                    }
-                    else if (valueClrType.Equals(clientTemplate.GuidTypeName)
-                        | valueClrType.Equals(clientTemplate.DateTimeOffsetTypeName)
-                        | valueClrType.Equals(clientTemplate.DateTypeName)
-                        | valueClrType.Equals(clientTemplate.TimeOfDayTypeName))
+            string valueClrType = GetClrTypeName(edmTypeReference, useDataServiceCollection, clientTemplate, context);
```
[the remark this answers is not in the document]
habbes: This code is auto-generated.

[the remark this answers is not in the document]
habbes: I changed in the source `ttinclude` file.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -3979,6 +3989,18 @@ public sealed class ODataClientCSharpTemplate : ODataClientTemplate
     } }
     private HashSet<string> CSharpKeywords;
 
+    internal override string StripNullableFromTypeName(string typeName)
+    {
+        const string nullablePrefix = "Nullable<";
+        int indexOfNullablePrefix = typeName.IndexOf(nullablePrefix);
```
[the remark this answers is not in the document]
habbes: This doesn't seem to be a rule that we adhere to in the codebase. I think we're _strict_ about empty lines after closing braces but not before if statements. In this case I think it makes sense to logically group the null check with the `indexOfNullablePrefix` definition because the variable is not safe to use without that null check, they go hand-in-hand.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// ------------------------------------------------------------------------------
 // <auto-generated>
 //     This code was generated by a tool.
-//     Runtime Version: 16.0.0.0
+//     Runtime Version: 17.0.0.0
```
ElizabethOkerio, line 4: will this change cause any issues?

habbes: No. It's only a comment. It's auto generated and indicates the version of VS templating tool. What counts is that the generated code compiles and runs on supported versions on .NET.

ElizabethOkerio (approved):

KenitoInc (approved):
LGTM

Issue #380, opened 2024-03-04
DynamicProperties property for open type entity not wrapped with ContainerProperty attribute


stanb:
### Describe the bug
I have service with open type entity:
```
    public class ResultObject
    {
        public string Id { get; set; } = string.Empty;
        public IDictionary<string, object> DynamicProperties { get; set; } = new Dictionary<string, object>();
    }
```
Metadata:
```xml
      <EntityType Name="ResultObject" OpenType="true">
        <Key>
          <PropertyRef Name="Id" />
        </Key>
        <Property Name="Id" Type="Edm.String" Nullable="false" />
      </EntityType>
```
ODataConnectedService generated class:
```cs
    [global::Microsoft.OData.Client.Key("Id")]
    [global::Microsoft.OData.Client.OriginalNameAttribute("ResultObject")]
    public partial class ResultObject : global::Microsoft.OData.Client.BaseEntityType
    {
        /// <summary>
        /// Create a new ResultObject object.
        /// </summary>
        /// <param name="ID">Initial value of Id.</param>
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "#VersionNumber#")]
        public static ResultObject CreateResultObject(string ID)
        {
            ResultObject resultObject = new ResultObject();
            resultObject.Id = ID;
            return resultObject;
        }
        /// <summary>
        /// There are no comments for Property Id in the schema.
        /// </summary>
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "#VersionNumber#")]

        [global::Microsoft.OData.Client.OriginalNameAttribute("Id")]
        [global::System.ComponentModel.DataAnnotations.RequiredAttribute(ErrorMessage = "Id is required.")]
        public virtual string Id
        {
            get
            {
                return this._Id;
            }
            set
            {
                this.OnIdChanging(value);
                this._Id = value;
                this.OnIdChanged();
            }
        }
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "#VersionNumber#")]
        private string _Id;
        partial void OnIdChanging(string value);
        partial void OnIdChanged();
        /// <summary>
        /// There are no comments for Property DynamicProperties in the schema.
        /// </summary>
[...]

PR #381, opened 2024-03-05
Fix missing ContainerProperty attribute on dynamic properties (issue #380)

stanb:
### Issues
This pull request fixes issue #380 
When entity is an open type (`<EntityType Name="ResultObject" OpenType="true">`) VS extension generates `IDictionary<string, object> DynamicProperties` property. 
This property must have `ContainerProperty` attribute. 
Otherwise, when connected service is used, the dictionary is not populated from response.
If in the project, referenced `Microsoft.OData.Client` nuget with version >= 7.10.*, `ContainerProperty` attribute not added due to wrong versions comparison.
If I reference `Microsoft.OData.Client` version 7.20.0, the `ContainerProperty` attribute not added because lexically 7.20.0 < 7.6.4
### Description
Fix `Microsoft.OData.Client` versions comparison.

### DIFF
--- a/src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs
+++ b/src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs
@@ -83,7 +83,7 @@ namespace Microsoft.OData.ConnectedService
                         // Assembly reference (For project reference, SourceProject != null)
                         if (reference.Name.Equals("Microsoft.OData.Client", StringComparison.Ordinal))
                         {
-                            return Version.Parse(reference.Version) > Version.Parse("7.6.4.0");
+                            return Version.Parse(reference.Version) >= Version.Parse("7.6.4.0");
                         }
                     }
                 }
--- a/src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs
+++ b/src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs
@@ -83,7 +83,7 @@ namespace Microsoft.OData.ConnectedService
                         // Assembly reference (For project reference, SourceProject != null)
                         if (reference.Name.Equals("Microsoft.OData.Client", StringComparison.Ordinal))
                         {
-                            return Version.Parse(reference.Version) >= Version.Parse("7.6.4.0");
+                            return Version.Parse(reference.Version) > Version.Parse("7.6.4.0");
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


stanb:

wachugamaina:
Could you kindly share in the description what the issue is exactly and what the solution hopes to achieve?

stanb:
@wachugamaina I hope this is enough.

habbes:
@stanb thanks for this PR. Using lexical comparison was definitely a strange choice. Would you mind rebasing this PR with the `master` branch?

@gathogojr could you check why the pipeline is failing?

stanb:
@habbes done

gathogojr:
Hey @stanb. Your code change won't give you the outcome you expect.
Microsoft.OData.Client version 7.6.4 didn't support the dynamic properties container property.
Support was introduced in 7.7.*
Take a look at the changelog here - https://learn.microsoft.com/en-us/odata/changelog/odatalib-7x#odatalib-770-beta-release
It's the reason why the version was explicitly hard-coded.
Your code would result into the dynamic properties container property being generated in the proxy but the logic to populate the dictionary with the materialized dynamic properties isn't available in 7.6.4 and lower versions of the library.

gathogojr:
> @stanb thanks for this PR. Using lexical comparison was definitely a strange choice. Would you mind rebasing this PR with the `master` branch?
> 
> @gathogojr could you check why the pipeline is failing? A fix for the failures on the build pipeline had been merged.

@habbes The rebase is what was needed. It had

stanb:
@gathogojr fixed

gathogojr (approved):

src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs
```diff
@@ -81,9 +81,9 @@ public bool EmitContainerPropertyAttribute()
                     if (reference.SourceProject == null)
                     {
                         // Assembly reference (For project reference, SourceProject != null)
-                        if (reference.Name.Equals("Microsoft.OData.Client", StringComparison.Ordinal) && string.Compare(reference.Version, "7.6.4.0", StringComparison.Ordinal) > 0)
+                        if (reference.Name.Equals("Microsoft.OData.Client", StringComparison.Ordinal))
                         {
-                            return true;
+                            return Version.Parse(reference.Version) > Version.Parse("7.6.4.0");
```
ElizabethOkerio, line 86: Can we have a parsing exception here? or reference.Version will always be in the required format?

gathogojr: Especially because the reference is for a library we maintain, plus this statement is in within the block
```csharp
if (reference.Name.Equals("Microsoft.OData.Client", StringComparison.Ordinal))
```
I think this should be safe.
In addition, we currently update the Microsoft.OData.Client reference manually.

habbes (approved):

PR #362, opened 2023-06-08
Add no timestamp option (issue #349)

gregwinterstein:
Add the NoTimestamp option to disable the Generated date timestamp in generated files.
Added to OData-CLI and to VS Extension
Tests added and updated to take into account NoTimestamp option

Fixes #349

PR #383, opened 2024-03-08
Support omitting runtime version and code generation timestamp from generated files

### LINKED ISSUE #379: Generated files contain a timestamp that is more harmful than helpful
Yes - the problem is that the file(s) generated include a comment showing the `Generation date` as a date/time. This makes it difficult to see what files have _really_ changed when re-running/updating the connected service, because _all_ files get changed at least to have an updated `Generation date`.

I would like it to either:

- a) not write the `Generation date` comment at all
- b) have an option to disable writing the `Generation date` comment

None.

#### Additional context
Here are the top few lines of a generated file that show the `Generation date` comment:

``` csharp
//------------------------------------------------------------------------------
// <auto-generated>
//     Runtime Version:4.0.30319.42000
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

// Generation date: 2/14/2024 5:49:22 PM
namespace MyServiceNamespace.Entities
{
```

gathogojr:
This pull request replaces https://github.com/OData/ODataConnectedService/pull/362 and fixes #379

It supports omitting runtime version and code generation timestamp from generated files. When the option is set to true, the auto-generated code disclaimer section will look as follows:
```csharp
//------------------------------------------------------------------------------
// <auto-generated>
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

```
as opposed to:
```csharp
//------------------------------------------------------------------------------
// <auto-generated>
//     Runtime Version:4.0.30319.42000
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

// Generation date: 08.03.2021 19:55:12
```

The `no-timestamp` / `ti` option introduced in #362 is replaced with `omit-versioning-info` / `vi`.

No release of OData Connected Service or CLI has been done subsequent to #362 being merged so having a common flag to control whether runtime version and code generation timestamp should omitted was deemed to be a better option.

### Additional Details
This pull request also fixes an issue causing the builds to fail by upgrading the VS build tools:
image

### DIFF
--- a/test/ODataConnectedService.Tests/ODataConnectedService.Tests.csproj
+++ b/test/ODataConnectedService.Tests/ODataConnectedService.Tests.csproj
@@ -111,10 +111,10 @@
       <Version>16.2.45</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VisualStudio.SDK">
-      <Version>17.3.32804.24</Version>
+      <Version>17.9.37000</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VSSDK.BuildTools">
-      <Version>17.3.2094</Version>
+      <Version>17.9.3168</Version>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
@@ -128,7 +128,7 @@
       <Version>1.3.2</Version>
     </PackageReference>
     <PackageReference Include="NuGet.VisualStudio">
-      <Version>17.6.0</Version>
+      <Version>17.9.1</Version>
     </PackageReference>
     <PackageReference Include="System.ComponentModel.Annotations">
       <Version>4.5.0</Version>


gathogojr:
@gregwinterstein @jphorv-bdo @rpallares FYI

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -95,10 +95,10 @@ public GenerateCommand()
 
             this.AddOption(internalModifier);
 
-            Option noTimestamp = new Option<bool>(new[] { "--no-timestamp", "-nt" })
+            Option noTimestamp = new Option<bool>(new[] { "--omit-versioning-info", "-vi" })
```
habbes, line 98: Would `-ovi` be a better shorthand?

habbes (approved):

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -397,9 +397,9 @@ public bool MakeTypesInternal
 }
 
 /// <summary>
-/// true to omit generation timestamp; otherwise false.
+/// true to omit runtime version and code generation timestamp; otherwise false.
```
ElizabethOkerio, line 400: nit: ```suggestion
/// Set to true to exclude the runtime version and code generation timestamp; otherwise, set to false.
```

File: src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs [hunk 3]
```diff
@@ -6312,24 +6322,28 @@ internal override void WriteFileHeader()
     {
 
 this.Write("\'------------------------------------------------------------------------------\r\n" +
-        "\' <auto-generated>\r\n\'     This code was generated by a tool.\r\n\'     Runtime Vers" +
-        "ion:");
+        "\' <auto-generated>\r\n\'     This code was generated by a tool.\r\n");
+
+
```
ElizabethOkerio, line 6327:
```suggestion
```

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -6312,24 +6322,28 @@ internal override void WriteFileHeader()
...
-'     the code is regenerated.
-' </auto-generated>
-'------------------------------------------------------------------------------
-
-Option Strict Off
-Option Explicit On
-");
+this.Write("\r\n");
+
+
```
ElizabethOkerio, line 6337:
```suggestion
```

ElizabethOkerio: I'm not sure about these spaces but I think they should be removed.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.tt
```diff
@@ -36,8 +36,8 @@ public static class Configuration
 	// thereby making them invisible outside the assembly
 	public const bool MakeTypesInternal = false;
 
-	// If set to true, generation timestamps will be left out of generated files
-	public const bool NoTimestamp = false;
+	// If set to true, runtime version and code generation timestamp will be omitted from the generated files
```
ElizabethOkerio, line 39: In some instances the code string is different. Is it possible to have the same code string for this property throughout the code files.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -8629,7 +8643,7 @@ public class ODataT4CodeGeneratorBase
         /// <summary>
         /// The string builder that generation-time code is using to assemble generated output
         /// </summary>
-        protected System.Text.StringBuilder GenerationEnvironment
+        public System.Text.StringBuilder GenerationEnvironment
```
KenitoInc, line 8646: Why this change?

KenitoInc (commented):
Just one question.
Otherwise LGTM

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.tt
gathogojr: @ElizabethOkerio I have addressed this in the following manner:
- For properties (with getter and setter), I followed the common convention for boolean properties and adopted the following text: `Gets or sets a value indicating whether to omit runtime version and code generation timestamp from the generated files.`
- For constants, I adopted the following text: `This flag indicates whether to omit runtime version and code generation timestamp from the generated files`
- For the UI label and CLI option, I adopted the following text: `Omit runtime version and code generation timestamp from the generated files.`

The texts reads the same tarting from the word "omit" but ahead of that I appended text that fits best in the particular usage

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
gathogojr: @ElizabethOkerio See my comment here https://github.com/OData/ODataConnectedService/pull/383#discussion_r1519322063

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
gathogojr: @ElizabethOkerio This file is generated via the "Run Custom Tool" option in Visual Studio against the `ODataT4CodeGenerator.tt` file. Here's how the code that generates this output looks in the `ODataT4CodeGenerator.ttinclude` file:
```csharp
// <auto-generated>
<#+
    if (!this.context.OmitVersioningInfo)
    {
#>
//     Runtime Version:<#= Environment.Version #>
<#+
    }
#>
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

<#+
    if (!this.context.OmitVersioningInfo)
    {
#>
// Generation date: <#= DateTime.Now.ToString(global::System.Globalization.CultureInfo.CurrentCulture) #>
<#+
    }
```
If we're to choose between keeping the logic in the `ttinclude` file legible plus easy to follow on the one side, and extra empty lines in the generated file, I'd rather the former.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
gathogojr: @ElizabethOkerio See my comment here https://github.com/OData/ODataConnectedService/pull/383#discussion_r1519337199

src/Microsoft.OData.Cli/GenerateCommand.cs
gathogojr: @habbes @KenitoInc We seem to have no specific convention at in respect to crafting the shorthand:
- `file-name` = `fn` --- I think this should have been just `f`. filename is a word...
- `outputdir` = `o` --- I find this okay. The shorter the better
- `metadata-uri` = `m` --- I find this okay too.
- `custom-headers` = `h` --- I find this okay too.
- `namespace-prefix` = `ns` --- Not `np` or `nsp`.  --- I find `ns` okay. Would have been fine with `n`.
- `upper-camel-case` = `ucc` --- IMO, a shorter form like `u` would have sufficed - so long as it's not being used as a short-hand for another option.
- `enable-internal` = `i` --- I find this okay too.
- `multiple-files` has no short-hand!!!
- `excluded-bound-operations` = `ebo` --- IMO, a shorter form like `eo` would have sufficed (i.e., `e` for exclude, while the second letter disambiguate between bound operations, operation imports and schema types).
- `excluded-operation-imports` = `eoi` --- IMO, a shorter form like `ei` would have sufficed.
- `excluded-schema-types` = `est` --- IMO, a shorter form like `et` would have sufficed.

The point I'm trying to illustrate with this is that we're not consistently combining the first letters from the option when crafting the shorthand. It's not a practice that I find common in any case. In my experience the shorthand of the option is seems to be typically kept as short as possible - one letter if possible, two letters to disambiguate.

ElizabethOkerio (approved):

KenitoInc (approved):

jphorv-bdo:
Thank you everyone for your work on this, I look forward to seeing it published!

jphorv-bdo:
> Thank you everyone for your work on this, I look forward to seeing it published!

Any estimate as to when this will be published to VS Marketplace?

jphorv-bdo:
@ElizabethOkerio @KenitoInc @habbes  I just realized there have been 38 PRs merged since the last time a release was published to VS Marketplace on August 10, 2022.

I wish I had looked at that earlier - if I had, I wouldn't have bothered checking for a new release every day for the past month!

When will a new release be published to VS Marketplace?

PR #388, opened 2024-04-11
Fix dependencies and code violations warnings

gathogojr:
This pull request fixes dependencies and code violations warnings

### DIFF
--- a/src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
+++ b/src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
@@ -53,8 +53,6 @@
     <Reference Include="Microsoft.Data.Services.Design">
       <HintPath>..\..\external\Binaries\V3\Microsoft.Data.Services.Design.dll</HintPath>
     </Reference>
-    <Reference Include="Microsoft.VisualStudio.TextTemplating.15.0, Version=16.0.28727, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
-    <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="PresentationCore" />
     <Reference Include="PresentationFramework" />
     <Reference Include="System" />
@@ -82,7 +80,6 @@
     <Compile Include="FileHandling\IFileHandler.cs" />
     <Compile Include="FileHandling\ODataFileOptions.cs" />
     <Compile Include="Logging\IMessageLogger.cs" />
-    <Compile Include="Logging\LogMessageCategory.cs" />
     <Compile Include="Logging\LogMessageCategory.cs" />
     <Compile Include="Models\BaseUserSettings.cs" />
     <Compile Include="PackageInstallation\IPackageInstaller.cs" />
@@ -138,6 +135,9 @@
     </PackageReference>
     <PackageReference Include="Microsoft.OData.Edm">
       <Version>7.6.3</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0">
+      <Version>10.0.30320</Version>
     </PackageReference>
     <PackageReference Include="Newtonsoft.Json">
       <Version>13.0.1</Version>
--- a/src/ODataConnectedService.Shared/Common/EdmHelper.cs
+++ b/src/ODataConnectedService.Shared/Common/EdmHelper.cs
@@ -16,7 +16,7 @@ using Microsoft.VisualStudio.ConnectedServices;
 
 namespace Microsoft.OData.ConnectedService.Common
 {
-    public class EdmHelper
+    internal static class EdmHelper
     {
         private static IDictionary<IEdmType, List<IEdmOperation>> _boundOperations = null;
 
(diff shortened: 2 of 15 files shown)


src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
```diff
@@ -139,6 +136,9 @@
     <PackageReference Include="Microsoft.OData.Edm">
       <Version>7.6.3</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.10.0">
```
ElizabethOkerio, line 139: Does this mean this dependency is no longer needed?

src/ODataConnectedService.Shared/Common/EdmHelper.cs
```diff
@@ -16,7 +16,7 @@
 
 namespace Microsoft.OData.ConnectedService.Common
 {
-    public class EdmHelper
+    internal static class EdmHelper
```
ElizabethOkerio, line 19: Will these be breaking changes?

src/ODataConnectedService.Shared/Common/UserSettingsPersistenceHelper.cs
```diff
@@ -134,7 +134,7 @@ private static IsolatedStorageFile GetIsolatedStorageFile()
             }
             catch (Exception ex)
             {
-                logger?.WriteMessageAsync(LoggerMessageCategory.Warning, failureMessage, failureMessageArg, ex);
+                logger?.WriteMessageAsync(LoggerMessageCategory.Warning, failureMessage, failureMessageArg, ex).Wait();
```
ElizabethOkerio, line 137: Is this blocking?

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
```diff
@@ -2104,15 +2104,6 @@ public void WriteEnumTypeWithMembers()
 
         #region Tests for WriteStructuredTypeDeclaration
 
-        private const string ComplexTypeEdmx = @"<?xml version=""1.0"" encoding=""utf-8""?>
```
ElizabethOkerio, line 2107: This is not used anywhere?

src/ODataConnectedService_VS2022Plus/ODataConnectedService_VS2022Plus.csproj
```diff
@@ -98,6 +98,9 @@
     <PackageReference Include="Microsoft.VisualStudio.SDK">
       <Version>17.3.32804.24</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.VisualStudio.Shell.15.0">
```
ElizabethOkerio, line 101: This too is not used anywhere?

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
gathogojr: @ElizabethOkerio The dependency is needed. Before it was being referenced from GAC:
image

The dependency I have gotten rid of is this one:
image
It was being referenced from GAC but at some point it was removed such that it was no longer a valid reference. But the project would still build since it wasn't required. The `Microsoft.VisualStudio.TextTemplating.Interfaces.10.0` reference was sufficient.

test/ODataConnectedService.Tests/Templates/ODataClientTemplateUnitTests.cs
gathogojr: Correct. It was just causing a warning

src/ODataConnectedService.Shared/Common/UserSettingsPersistenceHelper.cs
gathogojr: Yeah. If we don't call `.Wait()`, the log won't be written. We do that in quite a few places so there are quite a lot of warnings cause of possible deadlock. We should address that wholesomely in a future pull request.

src/ODataConnectedService_VS2022Plus/ODataConnectedService_VS2022Plus.csproj
gathogojr: In the absense of this reference, the dependency was being auto-resolved with a warning. I added the dependency explicitly to resolve the warning

src/ODataConnectedService.Shared/Common/EdmHelper.cs
gathogojr: These are helper classes that should never have been public in the first place. They should have been static too since all the methods are static. Due to the highly-opinionated nature of the logic in the helper classes, I'd find it very hard to believe that anyone would be taking a dependency on them.

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
habbes: Should we be concerned that we have a dependency on Visual Studio components from the core `CodeGen` library which is also used in the CLI?

src/ODataConnectedService.Shared/Common/EdmHelper.cs
habbes: We don't expose these assemblies are libraries, right? In that case, I don't think we should be that worried about breaking API changes.

src/Microsoft.OData.CodeGen/Microsoft.OData.CodeGen.csproj
gathogojr: @habbes I'm not 100% `Microsoft.VisualStudio.TextTemplating.Interfaces.10.0` dependency relies on Visual Studio per se. I believe the CLI needs it for text templating.

src/ODataConnectedService.Shared/Common/EdmHelper.cs
gathogojr: @habbes True

ElizabethOkerio (approved):

habbes (approved):

PR #389, opened 2024-04-18
Add service-name command line option

### LINKED ISSUE #384: Passing ServiceName as command line argument
Metadata file from odata.cli tool is always named `OData ServiceCsdl.xml`.
If you generate two references, the assembly will contain two resources with the same name ending, which will throw an exception in `GeneratedEdmModel.CreateXmlReader()`

Add optional --service-name (-sn) argument to the `generate` command to override the DefaultServiceName constant

I need to manually change resource name in Reference.cs and OData ServiceCsdl.xml filename

#### Additional context

ellizio:
Add service-name command line option
Resolves #384


ellizio:

PR #392, opened 2024-05-06
Fix broken command line options

### LINKED ISSUE #384: Passing ServiceName as command line argument

### LINKED ISSUE #390: Microsoft.OData.Cli 0.3.0 enable-tracking no longer working
There seems to be a regression when upgrading Microsoft.OData.Cli from 0.2.0 to 0.3.0.
Using an unmodified metadata.xml the new version no longer generate property tracking when the `--enable-tracking` flag is used.

### Assemblies affected

Microsoft.OData.Cli 0.3.0

### Reproduce steps

Generate client using odata-cli version 0.3.0 and the `--enable-tracking` flag.

### Expected result

Generated entities shall implement `INotifyPropertyChanged`.

### Actual result

Generated entities does **not** implement `INotifyPropertyChanged`.

### LINKED ISSUE #391: Microsoft.OData.Cli 0.3.0 --multiple-files no longer working
There seems to be a regression when upgrading Microsoft.OData.Cli from 0.2.0 to 0.3.0.
Using an unmodified metadata.xml the new version no longer generate multiple files when the --multiple-files flag is used.

Microsoft.OData.Cli 0.3.0

#### To Reproduce
1. Generate client using odata-cli version 0.3.0 and the --multiple-files flag
2. Observe the client is generated in one unique file

#### Expected behavior
The client shall be generated with multiple files

#### Actual behavior
The client is generated into one unique file that contains all classes

#### Additional context
The root cause is probably similar to https://github.com/OData/ODataConnectedService/issues/390 where some command options look ignored.

gathogojr:
This pull request fixes #390, fixes #391, fixes #384, and closes https://github.com/OData/ODataConnectedService/pull/389

This appears to have broken when introducing a feature that allows loading the command line options from a config file. Here's the pull request https://github.com/OData/ODataConnectedService/pull/363

It's a useful feature since it helps someone avoid having to specify the command line options all the time.

For the feature to work in a manner that makes logical sense, the value of a command line option that is explicitly specified should take precedence over the value loaded from the config file.

_There was one challenge with this though..._

Let's say the value for a particular option is `true` in the config file, **how would we differentiate between `false` that is explicitly specified from the command line (such that it should take precedence over the config file `true` option) if an unspecified option also defaults to `false`?**
This is the reason we changed the boolean options to be nullable such that their values would be null when not specified.

Users should still to be able to specify boolean options without an explicit true, e.g., `--enable-tracking` as opposed to `--enable-tracking true`. However, by making the options nullable, that capability broke. When a boolean option is not explicitly set to `true` or `false`, it's defaulting to `null`.

The workaround currently is to explicitly specify `true` or `false`, e.g., `--enable-tracking true` or `--enable-tracking false`.

I have since discovered that we could have achieved the objective without making the options nullable from the `GenerateCommand` class. We only need to make the property nullable from the `GenerateOptions`, i.e.:
```csharp
public class GenerateCommand : Command
{
    public GenerateCommand()
        : base("generate", "...")
    {
        // ... other options
        Option enableTracking = new Option<bool>(new[] { "--enable-tracking", "-et" })
        {
            Name = "enable-tracking",
            Description = "Enable entity and property tracking."
        }
        // ... other options
    }
}

public class GenerateOptions
{
    // ... other properties
    public bool? EnableTracking { get; set; }
    // ... other properties
}
```
This resolves the issue. For example, with the above change, both `--enable-tracking` and `--enable-tracking true` initialize the `EnableTracking` property to `true`, while unspecified options default to `null`.

From my investigation, I also discovered that other users of the **System.CommandLine** library had made a feature request for command line options to support three states - undefined|null/true/false: https://github.com/commandlineparser/commandline/issues/316
That feature has already been implemented and upgrading the CLI project to the latest version of the **System.CommandLine** library (without making any change) would also solve the issue.

This pull request resolves the reported issue by removing the `?` operator from the options in `GenerateCommand` class. It also upgrades the **System.CommandLine** library to the latest version.

### Additional work
- Added tests for the CLI command options and code generation
- Add support for `ServiceName` command line option to the CLI - This makes it possible to control the CSDL filename (Created as "Service NameCsdl.xml" by default)


gathogojr:
@gregwinterstein FYI

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,10 +248,38 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
             }
 
-            var namespacePrefix = string.IsNullOrEmpty(generateOptions.NamespacePrefix) ? configUserSettings?.NamespacePrefix : generateOptions.NamespacePrefix;
-            var excludedSchemaTypes = generateOptions.ExcludedSchemaTypes?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(type => type.Trim()).ToList();
-            var excludedBoundOperations = generateOptions.ExcludedBoundOperations?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(operation => operation.Trim()).ToList();
-            var excludedOperationImports = generateOptions.ExcludedOperationImports?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(import => import.Trim()).ToList();
+            // Empty namespace prefix commandline option should indicate not to apply any namespace prefix
+            var useNamespacePrefix = false;
+            string namespacePrefix = null;
+            if (!(generateOptions.NamespacePrefix != null && generateOptions.NamespacePrefix.All(char.IsWhiteSpace)))
```
habbes, line 254: If this if statement a bit hard to parse because of the double negation (`!` and `!=`). I think the following would be easier to read:
```suggestion
            if (generateOptions.NamespacePrefix == null && !generateOptions.NamespacePrefix.All(char.IsWhiteSpace))
```
We check if it's null or all whitespace. Why do we need this check though?

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,10 +248,38 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            {
+                namespacePrefix = string.IsNullOrEmpty(generateOptions.NamespacePrefix) ? configUserSettings?.NamespacePrefix : generateOptions.NamespacePrefix;
```
habbes, line 256: Why do we check `string.IsNullOrEmpty` if the wrapping if-block already checks that the string is null or all whitespace (which includes empty).

habbes: I have the same question for all the similar changes to the code blocks below.

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,10 +248,38 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            // Empty excluded bound operations commandline option should indicate not to have any bound operations excluded
+            List<string> excludedBoundOperations = null;
+            if (!(generateOptions.ExcludedBoundOperations != null && generateOptions.ExcludedBoundOperations.All(char.IsWhiteSpace)))
+            {
+                excludedBoundOperations = generateOptions.ExcludedBoundOperations?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(operation => operation.Trim()).ToList();
+                excludedBoundOperations = excludedBoundOperations?.Count > 0 ? excludedBoundOperations : configUserSettings?.ExcludedBoundOperations;
+            }
+
+            // Empty excluded operation imports commandline option should indicate not to have any operation imports excluded
+            List<string> excludedOperationImports = null;
```
habbes, line 277: See: https://github.com/OData/ODataConnectedService/pull/392/files#r1594421208

test/Microsoft.OData.Cli.Tests/CodeGeneration/CodeVerificationHelper.cs
```diff
@@ -0,0 +1,153 @@
+﻿//-----------------------------------------------------------------------------------
```
habbes, line 1: These seem like utilities we used in testing the connected service. Would it be beneficial to have a common testing utility project? Not a blocker, just a question.

test/Microsoft.OData.Cli.Tests/CodeGeneration/ODataCliCodeGenerationTests.cs
```diff
@@ -0,0 +1,475 @@
...
+        private readonly GenerateCommand generateCommand;
+        private static readonly string metadataUri;
+        private static readonly string lowerCamelCaseMetadataUri;
+        private static readonly string outputDir;
+
+        static ODataCliCodeGenerationTests()
+        {
+            metadataUri = Path.Combine(Environment.CurrentDirectory, "CodeGeneration\\Artifacts\\SampleServiceV4.xml");
+            lowerCamelCaseMetadataUri = Path.Combine(Environment.CurrentDirectory, "CodeGeneration\\Artifacts\\SampleServiceV4LowerCamelCase.xml");
+            outputDir = Path.Combine(Path.GetTempPath(), "twuf0gqr.d25");
```
habbes, line 24: What does `twuf0gqr.d25` refer to?

test/Microsoft.OData.Cli.Tests/CodeGeneration/ODataCliCodeGenerationTests.cs
```diff
@@ -0,0 +1,475 @@
...
+        }
+
+        public ODataCliCodeGenerationTests()
+        {
+            this.generateCommand = new GenerateCommand();
+            if (Directory.Exists(outputDir))
+            {
+                // Empty directory
```
habbes, line 32: It wasn't clear that this a verb and not an adjective, which was confusing for me at first.
```suggestion
                // Let's empty directory
```

test/Microsoft.OData.Cli.Tests/CodeGeneration/ODataCliCodeGenerationTests.cs
```diff
@@ -0,0 +1,475 @@
...
+                    foreach (var file in outputDirInfo.GetFiles())
+                    {
+                        file.Delete();
+                    }
+                }
+                catch (IOException)
+                {
+                    // Non-empty output directory could affect test outcomes?
+                    Assert.Fail("Non-empty output directory could affect test outcomes");
+                }
```
habbes, line 46: What if there are subfolders as well? Anyway, why not use `Directory.Delete(outputDir, true)` which deletes a directory recursively?: https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.delete?view=net-8.0#system-io-directory-delete(system-string-system-boolean)

habbes:
@gathogojr thanks for adding the tests. Do they run in the build pipeline as well?

test/Microsoft.OData.Cli.Tests/CodeGeneration/ODataCliCodeGenerationTests.cs
gathogojr: @habbes I changed to `Directory.Delete(outputDir, true)`. Also made the test class implement `IDisposable` and moved the logic to the `Dispose` method such that a temporary folder with a random name is created for each test and disposed after the test completes.

test/Microsoft.OData.Cli.Tests/CodeGeneration/ODataCliCodeGenerationTests.cs
gathogojr: It's just a random temp filename I had generated and was using during development. I changed the logic to:
```csharp
Path.Combine(Path.GetTempPath(), Path.GetRandomFileName())
```
The temp folder is delete from the `Dispose` method.

test/Microsoft.OData.Cli.Tests/CodeGeneration/CodeVerificationHelper.cs
gathogojr: @habbes In **ODataConnectedService.Tests**, we use **Microsoft.VisualStudio.TestPlatform.TestFramework** (and maybe **xunit** as well since the dependency is referenced) while it **Microsoft.OData.Cli.Tests** I decided to use **xunit** (that we use heavily in other OData repos).
There are test assertions done from the `CodeGenerationHelpers` class in **ODataConnectedService.Tests**.
Those assertions are important since they indicate where to code fragments differ.
They however make it tricky to move the helper class in a "TestCommon" project since the two test projects are not using the same test framework.
I do however that we should do everything possible to avoid code duplication. I'll create an issue to track the effort required to have the two test projects share the verification logic in the helper.

src/Microsoft.OData.Cli/GenerateCommand.cs
gathogojr: @habbes Like I explained offline, when either of `namespace-prefix`, `excluded-operation-imports`, `excluded-bound-operations`, and `excluded-schema-types` is explicitly specified from the command line as an empty string, that should interpreted as an indication that if there was a non-empty value specified previously undone, it should be undone. Otherwise, one would never be able to undo any namespace prefix, or excluded model elements from the config file.
As an illustration, consider a scenario, where on the first occasion I elected to have a namespace prefix "Client":
```
generate --metadata-uri http://tempuri.org --outputdir C:\tempdir --namespace-prefix Client
```
You end up with a config file with the following properties (among others):
```json
{
    "Endpoint": "http://tempuri.org",
    "UseNamespacePrefix": true,
    "NamespacePrefix": "Client"
}
```
The only way I'd be able to unset the ClientNS namespace prefix from the config file is as follows:
```
generate --metadata-uri http://tempuri.org --outputdir C:\tempdir --namespace-prefix "" --config-file C:\tempdir\ConnectedService.json
```
Where the empty string will be interpreted to mean I want to overwrite the value in the config file.

Same case with excluded model element. Consider a scenario, where on the first occasion I elected to exclude "ResetDataSource" operation import:
```
generate --metadata-uri http://tempuri.org --outputdir C:\tempdir --excluded-operation-imports ResetDataSource
```
You end up with a config file with the following properties (among others):
```json
{
    "Endpoint": "http://tempuri.org",
    "ExcludedOperationImports": ["ResetDataSource"]
}
```
The only way I'd be able to unset the ResetDataSource excluded operation import from the config file is as follows:
```
generate --metadata-uri http://tempuri.org --outputdir C:\tempdir --excluded-operation-imports "" --config-file C:\tempdir\ConnectedService.json
```

src/Microsoft.OData.Cli/GenerateCommand.cs
gathogojr: @habbes I rewrote the logic so it's much easier to follow. There's also the comment indicating what the logic is trying to accomplish

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,10 +248,38 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            string namespacePrefix = null;
+            if (!(generateOptions.NamespacePrefix != null && generateOptions.NamespacePrefix.All(char.IsWhiteSpace)))
+            {
+                namespacePrefix = string.IsNullOrEmpty(generateOptions.NamespacePrefix) ? configUserSettings?.NamespacePrefix : generateOptions.NamespacePrefix;
+                useNamespacePrefix = (configUserSettings?.UseNamespacePrefix ?? false) || (!string.IsNullOrWhiteSpace(namespacePrefix));
+            }
+
+            // Empty excluded schema types commandline option should indicate not to have any schema types excluded
+            List<string> excludedSchemaTypes = null;
+            if (!(generateOptions.ExcludedSchemaTypes != null && generateOptions.ExcludedSchemaTypes.All(char.IsWhiteSpace)))
```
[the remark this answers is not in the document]
gathogojr: @habbes I rewrote the logic so it's easier to follow. In addition, see my comment here https://github.com/OData/ODataConnectedService/pull/392#discussion_r1595194879

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,10 +248,38 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            {
+                excludedSchemaTypes = generateOptions.ExcludedSchemaTypes?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(type => type.Trim()).ToList();
+                excludedSchemaTypes = excludedSchemaTypes?.Count > 0 ? excludedSchemaTypes : configUserSettings?.ExcludedSchemaTypes;
+            }
+
+            // Empty excluded bound operations commandline option should indicate not to have any bound operations excluded
+            List<string> excludedBoundOperations = null;
+            if (!(generateOptions.ExcludedBoundOperations != null && generateOptions.ExcludedBoundOperations.All(char.IsWhiteSpace)))
```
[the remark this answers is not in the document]
gathogojr: @habbes I rewrote the logic so it's easier to follow. In addition, see my comment here https://github.com/OData/ODataConnectedService/pull/392#discussion_r1595194879

src/Microsoft.OData.Cli/GenerateCommand.cs
gathogojr: @habbes I rewrote the logic so it's easier to follow. In addition, see my comment here https://github.com/OData/ODataConnectedService/pull/392#discussion_r1595194879

gathogojr:
> @gathogojr thanks for adding the tests. Do they run in the build pipeline as well?

@habbes This was a good catch. I updated the build pipeline to run the CLI tests

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -91,48 +92,44 @@ public GenerateCommand()
 
             this.AddOption(ns);
 
-            Option enableTracking = new Option<bool?>(new[] { "--enable-tracking", "-et" })
+            Option enableTracking = new Option<bool>(new[] { "--enable-tracking", "-et" })
             {
                 Name = "enable-tracking",
-                Description = "Enable entity and property tracking."
+                Description = "Enable entity and property tracking.",
```
ElizabethOkerio, line 98: Is this comma needed here?

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -177,6 +173,14 @@ public GenerateCommand()
 
             this.AddOption(outputDir);
 
+            Option serviceName = new Option<string>(new[] { "--service-name", "-s" })
+            {
+                Name = "serviceName",
```
ElizabethOkerio, line 178:
```suggestion
                Name = "service-name", The other options have the long form identifier of the option as the name. For uniformity. 
```

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,15 +256,41 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
                 configUserSettings = configFileData?.ExtendedData;
             }
 
-            var namespacePrefix = string.IsNullOrEmpty(generateOptions.NamespacePrefix) ? configUserSettings?.NamespacePrefix : generateOptions.NamespacePrefix;
-            var excludedSchemaTypes = generateOptions.ExcludedSchemaTypes?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(type => type.Trim()).ToList();
-            var excludedBoundOperations = generateOptions.ExcludedBoundOperations?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(operation => operation.Trim()).ToList();
-            var excludedOperationImports = generateOptions.ExcludedOperationImports?.Split(",", StringSplitOptions.RemoveEmptyEntries).Select(import => import.Trim()).ToList();
+            // Empty namespace prefix commandline option should indicate not to apply any namespace prefix
```
ElizabethOkerio, line 259:
```suggestion
            // If the command line option for the namespace prefix is empty, then it signifies that no namespace prefix should be applied.
```

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,15 +256,41 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            string namespacePrefix = generateOptions.NamespacePrefix?.Trim();
+            if (namespacePrefix == null)
+            {
+                namespacePrefix = configUserSettings?.NamespacePrefix;
```
ElizabethOkerio, line 263: Up there. you state that if namespace prefix is null then no namespace prefix should be applied. Why are we applying the namespace prefix here? or What does `configUserSettings?.NamespacePrefix` contain

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -252,15 +256,41 @@ private TServiceConfig GetServiceConfiguration<TServiceConfig>(GenerateOptions g
...
+            }
+
+            // Empty excluded schema types commandline option should indicate not to have any schema types excluded
```
ElizabethOkerio, line 266: nit: These docs are not so clear. I don't know. You could leave them as they are but suggesting they be updated to: 

// If the command line option for excluded schema types is empty, then no schema types should be excluded.

src/Microsoft.OData.Cli/GenerateCommand.cs
habbes: I think the reasoning here is that the `namespacePrefix` could either be set through command option or config file. If it's not set by the command option, we check the config file, if it's no set there either, it would be null. But also, I believe @gathogojr is making a clear distinction between empty and null. If the `namespacePrefix` is an empty string, then the user has explicitly set it to an empty value, which would remove the current prefix if it exists.

habbes (approved):

ElizabethOkerio (approved):

PR #399, opened 2024-06-11
Onboard Sign CLI tool

gathogojr:
This pull request replaces the legacy SignClient tool for signing the VSIX and the CLI with the new Sign CLI tool
Sign CLI tool doesn't use client secret. It relies on the newly created "Code Signing" service connection.


src/ODataConnectedService_VS2022Plus/ODataConnectedService_VS2022Plus.csproj
```diff
@@ -99,7 +99,7 @@
       <Version>17.3.32804.24</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.VSSDK.BuildTools">
-      <Version>17.9.3168</Version>
```
gathogojr, line 102: Microsoft.VSSDK.BuildTools was updated to address the following failure on the pipeline:
> ##[error]C:\Users\VssAdministrator\.nuget\packages\microsoft.vssdk.buildtools\17.9.3168\tools\VSSDK\Microsoft.VsSDK.targets(895,5): Error MSB4018: The "GetDeploymentPathFromVsixManifest" task failed unexpectedly.

src/ODataConnectedService.Shared/ODataConnectedService.Shared.projitems
```diff
@@ -60,16 +60,4 @@
...
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="$(MSBuildThisFileDirectory)Resources\ExtensionIcon.png">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="$(MSBuildThisFileDirectory)Resources\Icon.png" />
-  </ItemGroup>
-  <ItemGroup>
-    <Resource Include="$(MSBuildThisFileDirectory)Resources\Icon.png" />
-  </ItemGroup>
```
gathogojr, line 74: The _License.txt_, _ExtensionIcon.png_, and _Icon.png_ referenced from the shared project were removed to cure the following failure on the pipeline:
> ##[error]C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.WinFx.targets(712,5): Error RG1000: Unknown build error, 'An item with the same key has already been added.' 
C:\Windows\Microsoft.NET\Framework\v4.0.30319\Microsoft.WinFx.targets(712,5): error RG1000: Unknown build error, 'An item with the same key has already been added.'  [D:\a\1\s\src\ODataConnectedService\ODataConnectedService.csproj]
Done Building Project "D:\a\1\s\src\ODataConnectedService\ODataConnectedService.csproj" (default targets) -- FAILED.

image

I don't believe it was necessary to reference those resources from the shared project since they're also referenced from both **ODataConnectedService** and **ODataConnectedService_VS2022Plus** projects directly from _src\SharedFiles_ directory.. They actually need to be referenced directly from both projects since they are added to the VSIX manifest file:
```
    <License>License.txt</License>
    <Icon>Resources\ExtensionIcon.png</Icon>
```
That's why it made sense to remove them from the shared project since they are not serving any purpose.

@ElizabethOkerio If am off the mark in my reasoning please let me know

src/ODataConnectedService.Shared/Resources/ExtensionIcon.png
gathogojr, line 1: See my comment here https://github.com/OData/ODataConnectedService/pull/399/files#r1635329757
It explains why this file was deleted

src/ODataConnectedService.Shared/Resources/Icon.png
gathogojr, line 1: See my comment here https://github.com/OData/ODataConnectedService/pull/399/files#r1635329757
It explains why this file was deleted

src/ODataConnectedService/source.extension.vsixmanifest
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="ODataConnectedService.e30335d6-f9c7-4d08-b66a-f011f3f18477" Version="1.1.1" Language="en-US" Publisher="Microsoft" />
```
gathogojr, line 4: Discussed about changing publisher to "Microsoft OData"

src/ODataConnectedService.Shared/License.txt
gathogojr, line 1: See my comment here https://github.com/OData/ODataConnectedService/pull/399/files#r1635329757
It explains why this file was deleted

azure-pipelines.yml
```diff
@@ -157,26 +157,31 @@ stages:
             inputs:
               command: custom
               custom: tool
-              arguments: install --tool-path . SignClient
+              arguments: install --tool-path . --prerelease sign
```
habbes, line 160: Why `--prerelease`?

azure-pipelines.yml
```diff
@@ -202,26 +207,31 @@ stages:
...
```
habbes, line 210: Why `--prerelease`

azure-pipelines.yml
```diff
@@ -260,24 +270,31 @@ stages:
...
```
habbes, line 273: Why `--prerelease`?

habbes (approved):
Verified that all the signed packages are signed

image

image

image

I wonder whether there's value in us continuing to package the non-signed packages to the pipeline artifacts, seems like it increases the surface area of making mistakes (e.g. deploying the wrong version of the package)?

azure-pipelines.yml
gathogojr: Only pre-release versions of the tool are available and one has to explicitly specify they're good with pre-release version otherwise the install command will fail.

azure-pipelines.yml
gathogojr: Only pre-release versions of the tool are available and one has to explicitly specify they're good with pre-release version otherwise the install command will fail.

azure-pipelines.yml
gathogojr: Only pre-release versions of the tool are available and one has to explicitly specify they're good with pre-release version otherwise the install command will fail.

src/ODataConnectedService.Shared/ODataConnectedService.Shared.projitems
ElizabethOkerio: I think the shared resources should be in the shared project then referenced in the individual vsix projects. Look at this doc:https://learn.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022#use-shared-projects-for-multi-targeting specifically point 6.

ElizabethOkerio:
> Verified that all the signed packages are signed
> 
> image
> 
> image
> 
> image
> 
> I wonder whether there's value in us continuing to package the non-signed packages to the pipeline artifacts, seems like it increases the surface area of making mistakes (e.g. deploying the wrong version of the package)?

How come the icon is not showing here?

src/ODataConnectedService.Shared/ODataConnectedService.Shared.projitems
ElizabethOkerio: I think they should be referenced from the shared directory.

gathogojr:
> > Verified that all the signed packages are signed
> > image
> > image
> > image
> > I wonder whether there's value in us continuing to package the non-signed packages to the pipeline artifacts, seems like it increases the surface area of making mistakes (e.g. deploying the wrong version of the package)?
> 
> How come the icon is not showing here or it normally doesn't show.

It appears like that even in old versions of the VSIX

src/ODataConnectedService.Shared/ODataConnectedService.Shared.projitems
gathogojr: @ElizabethOkerio In **ODataConnectedService** and **ODataConnectedService_VS2022Plus** projects, they are being referenced from a shared directory -> _/src/SharedFiles_.

In trying to resolve the error, I tried leaving only the references in the shared project and when I did that, the solution couldn't compile since the references to _**License.txt**_ and _**Resources\ExtensionIcon.png**_ in [_source.extension.vsixmanifest_](https://github.com/OData/ODataConnectedService/blob/7c5e685eb9d1df57ab8e13b3112a2e802cb6f722/src/ODataConnectedService/source.extension.vsixmanifest#L9) file expect the resources to be referenced directly in the respective projects.

For **ODataConnectedService** project:
https://github.com/OData/ODataConnectedService/blob/7c5e685eb9d1df57ab8e13b3112a2e802cb6f722/src/ODataConnectedService/ODataConnectedService.csproj#L166-L180

For **ODataConnectedService_VS2022Plus** project:
https://github.com/OData/ODataConnectedService/blob/7c5e685eb9d1df57ab8e13b3112a2e802cb6f722/src/ODataConnectedService_VS2022Plus/ODataConnectedService_VS2022Plus.csproj#L113-L127

The references in _**ODataConnectedService.Shared.projectitems**_ (the shared project) therefore don't seem to serve any purpose. In addition, if you observe you'll notice that _Icon.png_ is referenced two times as both `Content` and `Resource`:
https://github.com/OData/ODataConnectedService/blob/7c5e685eb9d1df57ab8e13b3112a2e802cb6f722/src/ODataConnectedService.Shared/ODataConnectedService.Shared.projitems#L63-L74

gathogojr: @ElizabethOkerio This change is also supported by the link you shared:
> Move metadata files (for example, release notes, license, and icons) and VSCT files to a shared directory. Then add them as linked files to the VSIX project. Note that the shared directory is separate from the shared project.

PR #407, opened 2024-08-15
Resolve conflicts when two csdl names have similar suffixes

### LINKED ISSUE #405: Parsing for the xml file name can result in exception with multiple services
### Describe the bug
If you have multiple services in your project [this here](https://github.com/OData/ODataConnectedService/blob/55a4768c6fc247fdc7bbe919555152868713b8d1/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs#L5124) can lead to a linq enumeration exception. (same goes for the ttinclude as well of course)

This should not be done via `EndsWith` but comparing the **full filename**.

ODataConnectedService latest
VisualStudio Version v17.10.4
Microsoft.Odata.Client v7.21.3

#### To Reproduce
1. Install a odata service, name it "OData Service"
2. Install another odata service and name it "My secondary OData Service" into the same assembly
3. make some boilerplate code to initialize and connect to the first service
4. Run
5. See error

#### Expected behavior
Expected behavior: 
Both services should be found, no exception should be thrown.

#### Actual behavior
Since both services end with "OData Service", the line mentioned above results in finding the xml for BOTH services.
So the comparison should not be done via EndsWith, it should check the whole file name, not just a part of it.

#### Additional context
Exception thrown:
`System.InvalidOperationException: 'Sequence contains more than one matching element'`

habbes:
Fixes #405 

We store the CSDL file as an embedded resource. When we want to load the and parse the CSDL in memory, we find the resource that matches the target CSDL file name. We use `resourceName.EndsWith(filePath)` to find the match. However, this causes a conflict when one CSDL name is a substring of the other, e.g.:

- `OData ServiceCsdl.xml`
- `Another OData ServiceCsdl.xml`

Currently an exception is thrown when such a conflict is detected. This PR fixes this issue by resolving the conflict by selecting the shortest resource name that ends with the target filename.

This solves the problem for simple cases, and possibly all practical common scenarios. However, it would be problematic in the following cases:
- The user manually creates a file called `OData ServiceCsdl.xml` in the same project and adds it as an embedded resource.
- The user renames the embedded resources or folders that were generated by the connected service or code generation tool.

I did not address the concerns because:
- Our code generation tool has the convention of adding the `Csdl.xml` suffix the service name set in the connected service configuration when generating the CSDL file. We do not expect the user to create files with this exact naming convention manually.
- We do not expect users to manually rename any files generated by the code generation tool. Doing so should be considered undefined behaviour.
- It would complicate the code.

The main issue here is that we `EndsWith` to match the files and resources. Why don't we just check the full resource name? The full resource name does not match the file name, it's generated based on the assembly namespace and connected service folder. It as the format: `<Namespace>.Connected_services.OData_Services.<filename>`. I am reluctant to generate a match based on the full resource name because I have not seen an official documentation of this resource naming convention. I am concerned about taking a dependency on a format that's not documented when there other easier alternatives. I don't know if there's a guarantee that this naming convention would not be changed in some future version of Visual Studio or .NET without our knowledge.

### About the tests

Frist, you many notice that while I have updated a lot of generated C# client code in the tests, I have not updated any test generated VB client code. This is because all of the tests in `ODataConnectedServiceTests` test against code that includes the CSDL in the reference code as a string, instead of as a separate file that's compiled as an embedded resource. That code path is not affected by the change in this PR. Only that `OData.Cli.Tests` test this code path. But we don't have any CLI tests for Visual Basic.

That made me realize that the OData CLI does not actually support Visual Basic, despite the fact the documentation says it does. I've created a separate issue to track that: https://github.com/OData/ODataConnectedService/issues/408

Secondly, while we do have tests to verify that the generate code looks as expected and compiles without errors, we don't have tests to verify that the code we generate runs as expected. For example, testing that when you run the generated code with two schemas that have overlapping names does not throw an exception or that the correct schema is selected. We achieve that using manual testing, which is tedious and not enforced consistently. I've created a separate issue to track an E2E testing environment: https://github.com/OData/ODataConnectedService/issues/409

### DIFF
--- a/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
+++ b/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
@@ -5121,7 +5121,8 @@ this.Write(@""")]
                 try
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
-                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
+                    // If multiple resource names end with the file name, select the shortest one.
+                    var resourcePath = global::System.Linq.Enumerable.First(assembly.GetManifestResourceNames().OrderBy(n => n.Length), str => str.EndsWith(filePath));
                     global::System.IO.Stream stream = assembly.GetManifestResourceStream(resourcePath);
                     return global::System.Xml.XmlReader.Create(new global::System.IO.StreamReader(stream));
                 }
--- a/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
+++ b/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
@@ -5122,7 +5122,9 @@ this.Write(@""")]
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
                     // If multiple resource names end with the file name, select the shortest one.
-                    var resourcePath = global::System.Linq.Enumerable.First(assembly.GetManifestResourceNames().OrderBy(n => n.Length), str => str.EndsWith(filePath));
+                    var resourcePath = global::System.Linq.Enumerable.First(
+                        global::System.Linq.Enumerable.OrderBy(assembly.GetManifestResourceNames(), n => n.Length),
+                        str => str.EndsWith(filePath));
                     global::System.IO.Stream stream = assembly.GetManifestResourceStream(resourcePath);
                     return global::System.Xml.XmlReader.Create(new global::System.IO.StreamReader(stream));
                 }
--- a/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
+++ b/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
@@ -7300,7 +7300,11 @@ this.Write(@""")>  _
             Private Shared Function CreateXmlReader() As Global.System.Xml.XmlReader
                 Try
                     Dim assembly As Global.System.Reflection.Assembly = Global.System.Reflection.Assembly.GetExecutingAssembly()
-                    Dim resourcePath As Global.System.String = Global.System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), Function(str) str.EndsWith(filePath))
+                    ' If multiple resource names end with the file name, select the shortest one.
+                    Dim resourcePath As Global.System.String =
+                        Global.System.Linq.Enumerable.First(
+                            Global.System.Linq.Enumerable.OrderBy(assembly.GetManifestResourceNames(), Function(n) n.Length),
+                            Function(str) str.EndsWith(filePath))
                     Dim stream As Global.System.IO.Stream = assembly.GetManifestResourceStream(resourcePath)
                     Return Global.System.Xml.XmlReader.Create(New Global.System.IO.StreamReader(stream))
                 Catch e As Global.System.Xml.XmlException
--- a/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
+++ b/src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
@@ -5123,8 +5123,9 @@ this.Write(@""")]
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
                     // If multiple resource names end with the file name, select the shortest one.
                     var resourcePath = global::System.Linq.Enumerable.First(
-                        global::System.Linq.Enumerable.OrderBy(assembly.GetManifestResourceNames(), n => n.Length),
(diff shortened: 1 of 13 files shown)


src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
```diff
@@ -5121,7 +5121,10 @@ internal override void WriteGeneratedEdmModel(string escapedEdmxString)
                 try
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
-                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
+                    // If multiple resource names end with the file name, select the shortest one.
+                    var resourcePath = global::System.Linq.Enumerable.First(
```
WanjohiSammy, line 5125: What happened if no CSDL file is found, would the `First()` not throw an exception? Or this is expected.

test/Microsoft.OData.Cli.Tests/CodeGeneration/Artifacts/SampleServiceV4EnableInternalProxy.cs
```diff
@@ -574,7 +574,10 @@ private abstract class GeneratedEdmModel
                 try
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
-                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
+                    // If multiple resource names end with the file name, select the shortest one.
+                    var resourcePath = global::System.Linq.Enumerable.First(
```
WanjohiSammy, line 578: What about the scenario where the names are similar but different in case. Is it necessary to add `StringComparison.OrdinalIgnoreCase` in the `EndsWith()` to be case insensitive?
```cs
... str.EndsWith(filePath, StringComparison.OrdinalIgnoreCase)
```

test/Microsoft.OData.Cli.Tests/CodeGeneration/Artifacts/SampleServiceV4EnableTrackingProxy.cs
```diff
@@ -651,7 +651,10 @@ private abstract class GeneratedEdmModel
                 try
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
-                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
+                    // If multiple resource names end with the file name, select the shortest one.
+                    var resourcePath = global::System.Linq.Enumerable.First(
```
WanjohiSammy, line 655: See my previous comments. If possible, do the same to the rest of the same change.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.cs
habbes: Yes it would throw an exception, and yes this is expected. That's why this block is wrapped in a `try-catch`. The exception is handled here: https://github.com/OData/ODataConnectedService/pull/407/files#diff-c202ce47d06c7411cdaa6daec28763115d6db1a96991834741c3948c5a012cb2L5130

image

test/Microsoft.OData.Cli.Tests/CodeGeneration/Artifacts/SampleServiceV4EnableInternalProxy.cs
habbes: Check the comments in the PR description. This CSDL file is auto-generated by our tool. The only case where it would be a different case is if the user renamed it.

WanjohiSammy (approved):

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenerator.ttinclude
```diff
@@ -4683,7 +4683,10 @@ namespace <#= fullNamespace #>
                 try
                 {
                     var assembly = global::System.Reflection.Assembly.GetExecutingAssembly();
-                    var resourcePath = global::System.Linq.Enumerable.Single(assembly.GetManifestResourceNames(), str => str.EndsWith(filePath));
+                    // If multiple resource names end with the file name, select the shortest one.
+                    var resourcePath = global::System.Linq.Enumerable.First(
```
gathogojr, line 4687: I feel like we should apply the filter first (i.e., `.EndsWith(string)`), then order by length, then pick the first item
 I feel like that's more efficient. We'll calculate length for fewer candidates (most times 1):
```csharp
global::System.Linq.Enumerable.First(
    global::System.Linq.Enumerable.OrderBy(
        global::System.Linq.Enumerable.Where(
            assembly.GetManifestResourceNames(),
            str1 => str1.EndsWith(filePath)),
        str2 => str2.Length))
```

There's also the question whether it'd be safer to use `FirstOrDefault()`

gathogojr (approved):

PR #173, opened 2020-10-23
Fix issue preventing OData Connected Service from being listed in VS

habbes:
Fix #172 

The connected service could not be loaded by Visual Studio because the `Microsoft.OData.ConnectedService.dll` inside the package was not properly signed.

Running `sn.exe -v Microsoft.OData.ConnectedService.dll` reported that it was delay-signed assembly. This means that the assembly still needs to be signed, indicating that the signing tool used in the build pipeline was not correctly signing the assembly.

The fix consists in:
- Disabling delay-signing
- Replacing the Microsoft-owned strong name public key file with a new strong name key-pair (including both public and private key, see the note on .NET core projects here: https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named)
- Simplifying the build pipeline

Now running `sn.exe -v Microsoft.OData.ConnectedService.dll` reports that the assembly is valid.

PR #411, opened 2024-08-25
Remove sync calls and replace them with async calls.

marabooy:
Replaces Synchronous I/O with async I/O. 
~Note that some calls had to be re-ordered as it seems having async calls then updating UI's afterwards for some components will always result in exceptions. The cause may be due to the threading model being quite different from the windows SDK  in the tests.~

Seems this was not needed as ConfigureAwait(true) resolves these as it resumes the thread in the main thread. However the tests need to be updated to run using [STA mode].
(https://www.nuget.org/packages/Xunit.StaFact) 

Note though this improve some of the UI updates to be more responsive. There is a performance penalty that lies within these lines [here in ConnectedServiceFileHandler](https://github.com/OData/ODataConnectedService/blob/master/src/ODataConnectedService.Shared/ConnectedServiceFileHandler.cs#L44).

```csharp
       /// <summary>
        /// Adds a file to a target path.
        /// </summary>
        /// <param name="fileName">The name of the file</param>
        /// <param name="targetPath">The path target where you want to copy a file to </param>
        /// <param name="oDataFileOptions">The options to use when adding a file to a target path.</param>
        /// <returns>Returns the path to the file that was added</returns>
        public async Task<string> AddFileAsync(string fileName, string targetPath, ODataFileOptions oDataFileOptions)
        {
            if (oDataFileOptions != null)
            {
                return await this.Context.HandlerHelper.AddFileAsync(fileName, targetPath, new AddFileOptions { SuppressOverwritePrompt = oDataFileOptions.SuppressOverwritePrompt, OpenOnComplete = oDataFileOptions.OpenOnComplete }).ConfigureAwait(true);
            }
            else
            {
                return await this.Context.HandlerHelper.AddFileAsync(fileName, targetPath).ConfigureAwait(true);
            }
        }
```
  Benchmarking the code shows that this could end-up causing some freezes for when files are many some considerations should be past a certain threshold (e.g. 200 files) we switch ask the user to switch to a `Namespace` grouping for better performance or the CLI tool. 

 ~This ` this.Context.HandlerHelper.AddFileAsync` API on multiple runs averaged 40 file additions per minute.~
_**I need to re-run this benchmark again as this seems to not be the case anymore Seems there are some random variations in the computer. I was running these tests that caused the slowdown earlier**_
  
 Another alternative would be to switch to the MSBuild same as with the CLI project or Project items (https://learn.microsoft.com/en-us/dotnet/api/envdte.projectitems?view=visualstudiosdk-2022) which in theory could provide better and faster API's.

Some quick benefits from this are
1. Filtering schema items is now faster and causes less re-draws since we only paint one page at a time.
2. Rendering schema items renders a paginated list which has less items hence the draw operation is fast.
3. Model is cached between pages from Config OData endpoint page to the final page reducing IO during page transitions.
4. The Async handlers removes duplicated code which overrode the current implementation with a similar method.
5. During generation of multiple files, we no-longer read existing files to determine if they are different, we truncate and overwrite the generated file. instead.) This reduces the IO operations and CPU operations. Note since these are done in the temp directory and not in the VS directory the benefit of checking if the file is similar is marginal.

Some future improvements will be:-
1. Compare generated files with existing project files before adding (note the read could be expensive).
2. Add a namespace grouping e.g. If the number of files exceed a certain number prompt the user to prefer namespace grouping instead based on
 $n= count(SchemaTypes)\ m = count(Namespaces)\ :then\ n >= m$ in almost all cases. So if $C=n/m$ we can always get a $C$ speedup while using namespace grouping.

### DIFF
--- a/src/Microsoft.OData.Cli/GenerateCommand.cs
+++ b/src/Microsoft.OData.Cli/GenerateCommand.cs
@@ -220,7 +220,8 @@ namespace Microsoft.OData.Cli
                 }
 
                 ServiceConfiguration config = GetServiceConfiguration<ServiceConfiguration>(options);
-                MetadataReader.ProcessServiceMetadata(config, out Version version);
+                config.Endpoint = MetadataReader.NormalizeUri(config);
+                var (_, version) = await MetadataReader.ProcessServiceMetadata(config);
                 if (version == Constants.EdmxVersion4)
                 {
                     await GenerateCodeForV4Clients(options, console).ConfigureAwait(false);
--- a/src/Microsoft.OData.Cli/GenerateCommand.cs
+++ b/src/Microsoft.OData.Cli/GenerateCommand.cs
@@ -220,8 +220,9 @@ namespace Microsoft.OData.Cli
                 }
 
                 ServiceConfiguration config = GetServiceConfiguration<ServiceConfiguration>(options);
-                config.Endpoint = MetadataReader.NormalizeUri(config);
-                var (_, version) = await MetadataReader.ProcessServiceMetadata(config);
+                config.Endpoint = MetadataReader.ValidateAndNormalizeUri(config.Endpoint);
+
+                var (_, version) = await MetadataReader.ProcessServiceMetadataAsync(config).ConfigureAwait(false);
                 if (version == Constants.EdmxVersion4)
                 {
                     await GenerateCodeForV4Clients(options, console).ConfigureAwait(false);
--- a/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
@@ -19,6 +19,7 @@ using Microsoft.OData.CodeGen.Models;
 using Microsoft.OData.CodeGen.PackageInstallation;
 using Microsoft.OData.CodeGen.Templates;
 
+
 namespace Microsoft.OData.CodeGen.CodeGeneration
 {
     public class V4CodeGenDescriptor : BaseCodeGenDescriptor
@@ -194,27 +195,38 @@ namespace Microsoft.OData.CodeGen.CodeGeneration
             await FileHandler.AddFileAsync(tempFile, metadataFile, new ODataFileOptions { SuppressOverwritePrompt = true });
 
             FileHandler.SetFileAsEmbeddedResource(csdlFileName);
+
             t4CodeGenerator.EmitContainerPropertyAttribute = FileHandler.EmitContainerPropertyAttribute();
 
             t4CodeGenerator.MetadataFilePath = metadataFile;
             t4CodeGenerator.MetadataFileRelativePath = csdlFileName;
 
+            await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 Generation Starting.");
+
+            var generationContent = await t4CodeGenerator.TransformTextAsync();
+
+            if (t4CodeGenerator.Errors != null && t4CodeGenerator.Errors.Count > 0)
+            {
+                foreach (var err in t4CodeGenerator.Errors)
+                {
+                    await MessageLogger.WriteMessageAsync(LogMessageCategory.Warning, err.ToString()).ConfigureAwait(false);
+                }
+            }
+
+
             using (StreamWriter writer = File.CreateText(tempFile))
             {
-                await writer.WriteAsync(t4CodeGenerator.TransformText());
+                await writer.WriteAsync(generationContent);
                 await writer.FlushAsync();
-                if (t4CodeGenerator.Errors != null && t4CodeGenerator.Errors.Count > 0)
-                {
-                    foreach (var err in t4CodeGenerator.Errors)
-                    {
-                        await MessageLogger.WriteMessageAsync(LogMessageCategory.Warning, err.ToString()).ConfigureAwait(false);
-                    }
-                }
             }
 
             var outputFile = Path.Combine(referenceFolder, $"{this.GeneratedFileNamePrefix(serviceConfiguration.GeneratedFileNamePrefix)}{(languageOption == LanguageOption.GenerateCSharpCode ? ".cs" : ".vb")}");
             await FileHandler.AddFileAsync(tempFile, outputFile, new ODataFileOptions { SuppressOverwritePrompt = true });
-            t4CodeGenerator.MultipleFilesManager?.GenerateFiles(serviceConfiguration.GenerateMultipleFiles, FileHandler, MessageLogger, referenceFolder, true, serviceConfiguration.OpenGeneratedFilesInIDE);
+
+            if (t4CodeGenerator.MultipleFilesManager != null)
+            {
+                await t4CodeGenerator.MultipleFilesManager?.CopyGeneratedFilesAsync(serviceConfiguration.GenerateMultipleFiles, FileHandler, MessageLogger, referenceFolder, fileCreated: true, serviceConfiguration.OpenGeneratedFilesInIDE);
+            }
             await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 was generated.");
         }
     }
--- a/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
@@ -216,18 +216,18 @@ namespace Microsoft.OData.CodeGen.CodeGeneration
 
             using (StreamWriter writer = File.CreateText(tempFile))
             {
-                await writer.WriteAsync(generationContent);
-                await writer.FlushAsync();
+                await writer.WriteAsync(generationContent).ConfigureAwait(false);
+                await writer.FlushAsync().ConfigureAwait(false);
             }
 
             var outputFile = Path.Combine(referenceFolder, $"{this.GeneratedFileNamePrefix(serviceConfiguration.GeneratedFileNamePrefix)}{(languageOption == LanguageOption.GenerateCSharpCode ? ".cs" : ".vb")}");
-            await FileHandler.AddFileAsync(tempFile, outputFile, new ODataFileOptions { SuppressOverwritePrompt = true });
+            await FileHandler.AddFileAsync(tempFile, outputFile, new ODataFileOptions { SuppressOverwritePrompt = true }).ConfigureAwait(false);
 
             if (t4CodeGenerator.MultipleFilesManager != null)
             {
                 await t4CodeGenerator.MultipleFilesManager?.CopyGeneratedFilesAsync(serviceConfiguration.GenerateMultipleFiles, FileHandler, MessageLogger, referenceFolder, fileCreated: true, serviceConfiguration.OpenGeneratedFilesInIDE);
             }
-            await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 was generated.");
+            await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 was generated.").ConfigureAwait(false);
         }
     }
 }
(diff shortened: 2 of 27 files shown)


marabooy:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.OData.Cli/GenerateCommand.cs
```diff
@@ -220,7 +220,8 @@ private async Task<int> HandleGenerateCommand(GenerateOptions options, IConsole
                 }
 
                 ServiceConfiguration config = GetServiceConfiguration<ServiceConfiguration>(options);
-                MetadataReader.ProcessServiceMetadata(config, out Version version);
+                config.Endpoint = MetadataReader.NormalizeUri(config);
+                var (_, version) = await MetadataReader.ProcessServiceMetadata(config);
```
WanjohiSammy, line 224: Add `ConfigureAwait(false)` on the `await`. This will ensure that no specific thread is waited and can use any thread available to continue with the operation.

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -194,27 +195,38 @@ private async Task AddGeneratedCodeAsync(string metadata, string outputDirectory
...
 
             FileHandler.SetFileAsEmbeddedResource(csdlFileName);
+
             t4CodeGenerator.EmitContainerPropertyAttribute = FileHandler.EmitContainerPropertyAttribute();
 
             t4CodeGenerator.MetadataFilePath = metadataFile;
             t4CodeGenerator.MetadataFileRelativePath = csdlFileName;
 
-            using (StreamWriter writer = File.CreateText(tempFile))
+            await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 Generation Starting.");
```
WanjohiSammy, line 204: Same here. Add `.ConfigureAwait(false)` to all the `await`

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -27,28 +28,11 @@ public static class MetadataReader
         /// <param name="serviceConfiguration">The <see cref="ServiceConfiguration"/> of the metadata provided.</param>
         /// <param name="edmxVersion">Edmx version of the metadata.</param>
         /// <returns>Location of the metadata file.</returns>
-        public static string ProcessServiceMetadata(ServiceConfiguration serviceConfiguration, out Version edmxVersion)
+        public static async Task<(string Location, Version EdmxVersion)> ProcessServiceMetadata(ServiceConfiguration serviceConfiguration)
```
ElizabethOkerio, line 31:
```suggestion
        public static async Task<(string Location, Version EdmxVersion)> ProcessServiceMetadataAsync(ServiceConfiguration serviceConfiguration)
```

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
```diff
@@ -118,19 +102,42 @@ public static string ProcessServiceMetadata(ServiceConfiguration serviceConfigur
...
+        {
+            if (string.IsNullOrEmpty(serviceConfiguration.Endpoint))
+            {
+                throw new ArgumentNullException("OData Service Endpoint", string.Format(CultureInfo.InvariantCulture, Constants.InputServiceEndpointMsg));
+            }
+
+            if (serviceConfiguration.Endpoint.StartsWith("https:", StringComparison.Ordinal)
+                || serviceConfiguration.Endpoint.StartsWith("http", StringComparison.Ordinal))
+            {
+                if (!Uri.TryCreate(serviceConfiguration.Endpoint, UriKind.Absolute, out Uri uri))
```
ElizabethOkerio, line 128: Could there be other issues causing this URI not to be created other than it not being an absolute URI? If so, I think this error message may not be sufficient to tell what causes this method to fail.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
ElizabethOkerio: Should we use ValueTask?

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -126,43 +130,51 @@ public class FilesManager {
     /// </summary>
     /// <param name="split">If true the function is executed and multiple files generated
     /// otherwise only a single file is generated.</param>
-    [SecurityCritical]
-    public virtual void GenerateFiles(bool split, IFileHandler handlerHelper, IMessageLogger logger, string referenceFolder, bool fileCreated, bool OpenGeneratedFilesInIDE)
+    public virtual async Task GenerateFiles(bool split)
```
ElizabethOkerio, line 133: nit: ```suggestion
    public virtual async Task GenerateFilesAsync(bool split)
```

**Comment by ElizabethOkerio** [comment]
Could you please provide more information on the calls that you've had to reorder and the UI updates that have been more responsive because of this.

File: src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenFilesManager.ttinclude [hunk 6]
```diff
@@ -172,12 +184,14 @@ public class FilesManager {
     /// </summary>
     /// <param name="fileName">Name of the file to be created</param>
     /// <param name="content">Content of the file to be created</param>
-    protected virtual void CreateFile(string fileName, string content)
+    protected virtual async Task CreateFileAsync(string fileName, string content)
```
WanjohiSammy, line 187: Can we keep the sync methods unchanged and simply add their async counterparts? This approach might help avoid breaking changes.

src/ODataConnectedService.Shared/Views/ConfigODataEndpoint.xaml.cs
```diff
@@ -55,7 +58,7 @@ private void OpenMetadataFileButton_OnClick(object sender, RoutedEventArgs e)
             }
         }
 
-        private void OpenConnectedServiceJsonFileButton_Click(object sender, RoutedEventArgs e)
+        private async void OpenConnectedServiceJsonFileButton_Click(object sender, RoutedEventArgs e)
```
WanjohiSammy, line 61: Avoid `async void`

Also if possible, create new `async` methods and leave the `sync` ones unchanged.

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTests.cs
```diff
@@ -285,7 +286,7 @@ public void CodeGen_DSVGreaterThanMDSV()
 ";
             try
             {
-                CodeGenWithT4Template(invalidEdmxDsvGreaterThanMdsv, null, true, false);
+                CodeGenWithT4TemplateAsync(invalidEdmxDsvGreaterThanMdsv, null, true, false);
```
WanjohiSammy, line 289: Add `await` and `.ConfigureAwait(false)`

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTest.cs
```diff
@@ -88,12 +89,12 @@ public void TestEntitiesComplexTypesEnumFunctionsDSCWithInternalTypes()
...
 
             };
-            var output = generator.TransformText();
+            var output = await generator.TransformTextAsync().ConfigureAwait(false);
             GeneratedCodeHelpers.VerifyGeneratedCode(expected, output);
         }
 
         [TestMethod]
-        public void TestEntitiesComplexTypesEnumFunctionsDSCWithNoVersioningInfo()
+        public async Task TestEntitiesComplexTypesEnumFunctionsDSCWithNoVersioningInfoAsync()
```
WanjohiSammy, line 97: We can keep sync tests and create new async tests. Refer to https://github.com/OData/ODataConnectedService/pull/411#discussion_r1732157759

Also have tests that ensure that the results of `sync` methods are equal to the results of corresponding `async` methods.

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTest.cs
```diff
@@ -121,12 +122,12 @@ public void TestEntitiesComplexTypesEnumFunctionsDSCWithInternalTypesWithNoVersi
...
                 OmitVersioningInfo = true,
             };
-            var output = generator.TransformText();
+            var output = await generator.TransformTextAsync().ConfigureAwait(false);
             GeneratedCodeHelpers.VerifyGeneratedCodeOmitVersioningInfo(expected, output);
         }
 
         [TestMethod]
-        public void TestTypeDefinitionsParamsConvertedToUnderlyingType()
+        public async Task TestTypeDefinitionsParamsConvertedToUnderlyingTypeAsync()
```
WanjohiSammy, line 130: Same comment for the rest of the tests:

[#173 (comment)](https://github.com/OData/ODataConnectedService/pull/411/files#r1732187879)

src/ODataConnectedService.Shared/Views/ConfigODataEndpoint.xaml.cs
marabooy: So for button clicks or event handlers the expectation is to use Async void else the handler does not allow for binding async Task. The alternative would be to move this code to another method that is called and leave the handler as a synchronous method

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
marabooy: Do we release this as a nuget package that people can use? If so we need to retain the method. I will leave a version of this accessible which routes to the async version. Since we await the return of get response from the webRequest I am not sure if the ValueTask will be correct to use here as there could be a number of exceptions that can arise.

test/ODataConnectedService.Tests/Templates/ODataT4CodeGeneratorTest.cs
marabooy: The current synchronous tests are not a good reflection of how the application runs and leaving them offers few benefits. The UI applications should run in STA threads that currently XUnit does not support unless when using libraries such as Xunit.STAFact. Since we don't use the libraries the test were written in a single threaded synchronous way which poorly reflects the runtime the applications run in. 

The options is to add STAFact or a similar capability to the testing infrastructure so the tests map to the runtime and update all tests to be async to match the current runtime.

src/Microsoft.OData.CodeGen/Templates/ODataT4CodeGenFilesManager.ttinclude
marabooy: Resolving this as the internal code in the file manager cannot be re-used through extending the class.

src/Microsoft.OData.CodeGen/Common/MetadataReader.cs
marabooy: I mostly refacored this to move it to a new block where the uri  side effects of being modified are more pronounced. 

To make the errors more apparent should I promote the error message and provide it to the user? This is because the number of factors that could lead to an unsuccessful url parsing could be many.

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
odero: Careful. You should not add `ConfigureAwait(false)` blindly to all `await` calls. This is a client application not a library

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -53,7 +54,7 @@ public override async Task AddGeneratedClientCodeAsync(string metadata, string o
             }
             else
             {
-                await AddGeneratedCodeAsync(metadata, outputDirectory, languageOption, serviceConfigurationV4);
+                await AddGeneratedCodeAsync(metadata, outputDirectory, languageOption, serviceConfigurationV4).ConfigureAwait(true);
```
odero, line 57: `.ConfigureAwait(true)` is the default await behaviour no need to explicitly add it

marabooy: Reverting this, I was trying to reproduce issue #412 as well as propose a new change within the shared code to help reduce impact of having to run some of the code outside the main thread.

marabooy:
Closing this as it has been broken down to smaller prs

PR #413, opened 2024-09-04
Fixes tests to use the correct context to enable async refactor

marabooy:
Refactors tests and prepares code for #411

Moves all UI tests to use await async and removes the .Wait() call that could hide any async bugs or COM extension communication bugs. Fixes #412 

Also introduces the use of Joinable task factory to switch to the background thread https://devblogs.microsoft.com/premier-developer/asynchronous-and-multithreaded-programming-within-vs-using-the-joinabletaskfactory/

PR #414, opened 2024-09-12
Migrate codegen to async

marabooy:
Updates codegen to be async and builds on top of #413 
- [x] Migrates the code generator to transform async.
- [x] Uses the Threading model within visual studio.
- [X] eliminates loading a file to see if it is different from the generated file.

### DIFF
--- a/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
+++ b/src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
@@ -201,7 +201,7 @@ namespace Microsoft.OData.CodeGen.CodeGeneration
 
             using (StreamWriter writer = File.CreateText(tempFile))
             {
-                await writer.WriteAsync(t4CodeGenerator.TransformText());
+                await writer.WriteAsync(await t4CodeGenerator.TransformTextAsync());
                 await writer.FlushAsync();
                 if (t4CodeGenerator.Errors != null && t4CodeGenerator.Errors.Count > 0)
                 {
@@ -214,7 +214,10 @@ namespace Microsoft.OData.CodeGen.CodeGeneration
 
             var outputFile = Path.Combine(referenceFolder, $"{this.GeneratedFileNamePrefix(serviceConfiguration.GeneratedFileNamePrefix)}{(languageOption == LanguageOption.GenerateCSharpCode ? ".cs" : ".vb")}");
             await FileHandler.AddFileAsync(tempFile, outputFile, new ODataFileOptions { SuppressOverwritePrompt = true });
-            t4CodeGenerator.MultipleFilesManager?.GenerateFiles(serviceConfiguration.GenerateMultipleFiles, FileHandler, MessageLogger, referenceFolder, true, serviceConfiguration.OpenGeneratedFilesInIDE);
+            if (t4CodeGenerator.MultipleFilesManager != null)
+            {
+               await t4CodeGenerator.MultipleFilesManager?.CopyGeneratedFilesAsync(serviceConfiguration.GenerateMultipleFiles, FileHandler, MessageLogger, referenceFolder, true, serviceConfiguration.OpenGeneratedFilesInIDE);
+            }
             await MessageLogger.WriteMessageAsync(LogMessageCategory.Information, "Client Proxy for OData V4 was generated.");
         }
     }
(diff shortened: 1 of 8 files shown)


test/ODataConnectedService.Tests/CodeGeneration/CodeGenDescriptorTest.cs
```diff
@@ -187,32 +188,27 @@ public void TestAddGenerateClientCode_GeneratesMultipleFiles()
...
-
             //The file manager expects the files to have been saved in the Temp directory
             // when ODataT4CodeGenerator.TransformText() was called. Since we're using a dummy code generator
             // we need to manually ensure those files exist
-            var file1TempPath = Path.Combine(Path.GetTempPath(), "File1.cs");
-            File.WriteAllText(file1TempPath, "Contents1");
-            var file2TempPath = Path.Combine(Path.GetTempPath(), "File2.cs");
-            File.WriteAllText(file2TempPath, "Contents2");
-
+            codeGen.MultipleFilesManager.GenerateFilesAsync(true).Wait();
```
habbes, line 207: Why not making this test async and use `async/await`? Or is that done in another PR?

habbes (commented):
Looks mostly good to me. I noticed some `Wait()` calls in the tests, is that intentional?

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -201,7 +201,7 @@ private async Task AddGeneratedCodeAsync(string metadata, string outputDirectory
 
             using (StreamWriter writer = File.CreateText(tempFile))
             {
-                await writer.WriteAsync(t4CodeGenerator.TransformText());
+                await writer.WriteAsync(await t4CodeGenerator.TransformTextAsync());
```
WanjohiSammy, line 204: By default, this will wait for the same thread to continue. Is that the expected behavior here

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
```diff
@@ -214,7 +214,10 @@ private async Task AddGeneratedCodeAsync(string metadata, string outputDirectory
 
             var outputFile = Path.Combine(referenceFolder, $"{this.GeneratedFileNamePrefix(serviceConfiguration.GeneratedFileNamePrefix)}{(languageOption == LanguageOption.GenerateCSharpCode ? ".cs" : ".vb")}");
```
WanjohiSammy, line 215: There has been recommendation from team member to use explicit types instead of `var`. I am not sure if that should also be applied to ODataConnectedService

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
marabooy: Yes this is the expected behaviour and the message handler provides users with a progress dialog when this method is executing.

src/Microsoft.OData.CodeGen/CodeGeneration/V4CodeGenDescriptor.cs
marabooy: This project unfortunately uses var in a lot of places and the reason in ODL var is discouraged is so we can catch any breaking changes. Since this is not shipped as a library there is no such concern.

E.g. In ODL 
```
IEnumerable<string> elements() => new string[]{....}

string[] elements() => new string[] {....} // This is flagged in ODL. However for locals we could use var in ODL without many issues but we maintain concrete types so that changes like these are flagged in all places where the function is called.
```

test/ODataConnectedService.Tests/CodeGeneration/CodeGenDescriptorTest.cs
marabooy: This is meant to be done in the STA pr. which this merges to 
image

marabooy: Let me resolve this in that pr as the changes will be quite long as this pr has a bunch of blocking wait calls.

WanjohiSammy (approved):

habbes (approved):

PR #415, opened 2024-09-24
UI fixes to allow smoother transitions between pages and paginate schema items

marabooy:
Breaks the UI code from the Async refactor in https://github.com/OData/ODataConnectedService/pull/411 

I added a simple paginator to improve the draw events on the schema page and operation imports page to reduce drawing overhead and increase how responsive the UI is.

The model since it is used severally in the Wizard the results are now cached after transitioning from the ConfigureODataModel page. So moving forward will ensure the UI responds smoothly eliminating IO.

with pagination.
image

Paginator with filter
image
image

### DIFF
--- a/src/ODataConnectedService.Shared/ODataConnectedServiceWizard.cs
+++ b/src/ODataConnectedService.Shared/ODataConnectedServiceWizard.cs
@@ -45,6 +45,11 @@ namespace Microsoft.OData.ConnectedService
 
         internal string ProcessedEndpointForSchemaTypes;
 
+        /// <summary>
+        /// An IEdmModel that allows us to reduce the number of I/O operations between page transitions after the ConfigureODataModel Page.
+        /// </summary>
+        internal IEdmModel Model { get; set; }
+
         public ODataConnectedServiceWizard(ConnectedServiceProviderContext context)
         {
             Context = context;
@@ -64,6 +69,7 @@ namespace Microsoft.OData.ConnectedService
             OperationImportsViewModel.PageEntering += OperationImportsViewModel_PageEntering;
             SchemaTypesViewModel.PageEntering += SchemaTypeSelectionViewModel_PageEntering;
             SchemaTypesViewModel.PageLeaving += SchemaTypeSelectionViewModel_PageLeaving;
+            ConfigODataEndpointViewModel.PageLeaving += ConfigODataEndpointViewModel_PageLeaving;
 
             if (Context != null && Context.IsUpdating)
             {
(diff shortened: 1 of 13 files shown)


habbes:
Could you share some screenshots to show the paginator?

src/ODataConnectedService.Shared/ODataConnectedServiceWizard.cs
```diff
@@ -175,6 +181,17 @@ public void ConfigODataEndpointViewModel_PageEntering(object sender, EventArgs a
             }
         }
 
+        public void ConfigODataEndpointViewModel_PageLeaving(object sender, EventArgs args)
+        {
+            if (sender is ConfigODataEndpointViewModel configOdataViewModel)
+            {
+                if (configOdataViewModel.View is ConfigODataEndpoint configOdataView)
+                {
+                    this.Model = EdmHelper.GetEdmModelFromFile(ConfigODataEndpointViewModel.MetadataTempPath);
```
habbes, line 190: Does `EdmHelper.GetEdmModelFromFile()` method perform I/O? Should this be async too? Should the model be cached here too?

src/ODataConnectedService.Shared/Views/OperationImports.xaml.cs
```diff
@@ -35,5 +27,53 @@ private void SelectAll_Click(object sender, RoutedEventArgs e)
         {
             (DataContext as OperationImportsViewModel).SelectAll();
         }
+
+        /// <summary>
+        /// Paginates data for schema types to allow the UI to render only one page of items at a time making the UI more responsive.
+        /// </summary>
+        /// <param name="pageNumber">The page to view indexes start from 1.</param>
+        public void DisplayPage(int pageNumber)
```
habbes, line 35: Do we have tests for this?

src/ODataConnectedService.Shared/Views/SchemaTypes.xaml.cs
```diff
@@ -68,5 +73,54 @@ private void DeselectAllBoundOperations_Click(object sender, RoutedEventArgs e)
         {
             (DataContext as SchemaTypesViewModel)?.DeselectAllBoundOperations();
         }
+
+        /// <summary>
+        /// Paginates data for schema types to allow the UI to render only one page of items at a time making the UI more responsive.
+        /// </summary>
+        /// <param name="pageNumber">The page to view indexes start from 1.</param>
+        public void DisplayPage(int pageNumber)
```
habbes, line 81: Seems like mostly the same logic from the `OperationsImport` control

habbes (approved):

src/ODataConnectedService.Shared/ODataConnectedServiceWizard.cs
marabooy: I have converted this to Async in the Async pr made this pr to be UI improvements without necessarily moving things to async as this will give us better experience without any changes.

src/ODataConnectedService.Shared/Views/OperationImports.xaml.cs
marabooy: Yes we do have tests for this. LoadOperationTypes_ShouldOnlyShowItemsInPaginator has a test that checks the bounds

src/ODataConnectedService.Shared/Views/SchemaTypes.xaml.cs
marabooy: I duplicated the logic into the User controls since I also didn't want to refactor the code to other areas since it only used in two places only and this would be a better faster way to get this to customers.

WanjohiSammy (approved):

PR #452, opened 2026-08-17
Omit optional request values from connected-service designer data

WanjohiSammy:
## Summary

Keep custom HTTP headers and web-proxy username/password values available for the active code-generation operation, but omit them from connected-service designer data. The default configuration path remains unchanged and avoids creating a copy when there are no optional values or legacy save flags to remove.

## Changes

- Mark custom HTTP headers and web-proxy username/password properties as ignored by DataContractSerializer, Newtonsoft.Json, and System.Text.Json.
- Persist a sanitized copy only when optional values or legacy save flags are present.
- Preserve the original in-memory configuration passed to code generation.
- Preserve non-optional proxy settings, including host and domain.
- Preserve the concrete `ServiceConfigurationV4` type and V4-specific values.
- Remove the two save controls because these values are no longer written to designer data.
- Capture serialized designer data in handler-path tests.

## Before

Enabling either save control retained the corresponding optional values on the configuration object passed to designer-data persistence.

## After

Designer data omits custom HTTP headers and web-proxy username/password values regardless of legacy save-flag values. The active generation operation still receives the original values. Configurations without those values or flags use the existing object directly.

### DIFF
--- a/src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
+++ b/src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.Serialization;
 
 namespace Microsoft.OData.CodeGen.Models
 {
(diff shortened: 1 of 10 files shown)


src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
```diff
@@ -22,12 +23,21 @@ public class ServiceConfiguration
         public bool MakeTypesInternal { get; set; }
         public bool OpenGeneratedFilesInIDE { get; set; }
         public bool GenerateMultipleFiles { get; set; }
+        [IgnoreDataMember]
+        [Newtonsoft.Json.JsonIgnore]
```
xuzhg, line 27: Do we can safely remove the newtonsoft.json?

src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
```diff
@@ -22,12 +23,21 @@ public class ServiceConfiguration
...
+        [System.Text.Json.Serialization.JsonIgnore]
         public string CustomHttpHeaders { get; set; }
         public bool IncludeWebProxy { get; set; }
         public string WebProxyHost { get; set; }
         public bool IncludeWebProxyNetworkCredentials { get; set; }
         public bool StoreWebProxyNetworkCredentials { get; set; }
+        [IgnoreDataMember]
+        [Newtonsoft.Json.JsonIgnore]
+        [System.Text.Json.Serialization.JsonIgnore]
```
xuzhg, line 36: do we still store the password in the code?

xuzhg (approved):

src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
WanjohiSammy: This is to prevent `Newtonsoft.Json (Json.NET)` from serializing or deserailizing this particular property

src/Microsoft.OData.CodeGen/Models/ServiceConfiguration.cs
WanjohiSammy: Maybe required at some instance when user wants to auth

